PR #43, opened 2017-08-31
DOJ-13: Changing the maximum character length for the name field.

Tom-Camp:


docroot/modules/custom/foia_personnel/foia_personnel.install
```diff
@@ -0,0 +1,27 @@
...
+    $spec = [
+      'type' => 'varchar',
+      'description' => '',
+      'length' => 255,
+      'not null' => FALSE,
+    ];
+    // Change field length to 255 characters.
+    $schema->changeField('foia_personnel', 'name', 'name', $spec);
+    $schema->changeField('foia_personnel_revision', 'name', 'name', $spec);
+  }
```
malikkotob: This looks great. Only comment is that since `foia_personnel_revision` is its own table, it would need its own conditional to check if the table exists. With that being said, if the `foia_personnel` table exists, chances are the revision one also exists since they're from the same module, so I'm not really too concerned.

PR #47, opened 2017-09-05
DOJ-8: Tests that request form template includes universal form fields.

dangur:


malikkotob:
Can we add a test that creates a new webform that uses the template and verifies that it has the fields from the template?

I'm also wondering if it would be difficult to use the template config to get the fields to be verified rather than hardcoding them. I'm guessing with behat not hardcoding them would be difficult, but some kind of automated tests against the `WebformTemplateController` in our custom `webform_template` module might be good.

PR #49, opened 2017-09-05
DOJ-9: Adding SimpleSamlPHP

Tom-Camp:


malikkotob:
Hi Tom, this looks good, but let's use the `blt` command to handle this setup, which does some acquia-specific configuration out of the box. Here is a link to the [docs](https://blt.readthedocs.io/en/8.x/readme/simplesamlphp-setup/).

Here are the high-level steps we'll be taking:
1. `blt simplesamlphp:init`
1. Add the same htaccess lines you did [here](https://github.com/usdoj/foia/pull/49/commits/83c686246a7c2d75b7dc827b13796e38921038f4)

While the docs cover taking further steps to configure the integration, we can hold off until we have our environments provisioned to avoid rework.

malikkotob:
Before running `blt simplesamlphp:init`, did you clear our the work you did in the first three commits? I ask because I still see the `simplesamlphp` library contents inside of the `simplesamlphp` directory in the root of the repo. I think BLT manages the library via composer in the vendor directory, so having the full library in the repo root isn't necessary. To my knowledge, it only uses that directory in the repo root to house unique configuration.

PR #56, opened 2017-09-08
DOJ-80: Updating core and dependencies.

Tom-Camp:


malikkotob:
Merged into `develop`.

PR #63, opened 2017-09-12
DOJ-94: Send request submission via email

Tom-Camp:


docroot/modules/custom/foia_webform/src/Plugin/WebformHandler/FoiaEmailWebformHandler.php
```diff
@@ -0,0 +1,134 @@
...
+   * Formats an array as a CSV string.
+   *
+   * @param array $data
+   *   The form submission data.
+   *
+   * @return array
+   *   Returns the array as a string.
+   */
+  public function arrayToString(array $data) {
+    $handle = fopen('php://temp', 'w');
```
dangur: @Tom-Camp For the second argument to fopen do we want to use c in lieu of w in case two temp files were writing at the same time? It looks like with c the first would lock the second out while with w the second would truncate the first.

Tom-Camp: Good catch! I'll make the change.

docroot/modules/custom/foia_webform/src/Plugin/WebformHandler/FoiaEmailWebformHandler.php
```diff
@@ -0,0 +1,134 @@
...
+        $error_message = 'No Submission Email: Unable to send email for %title';
+        $context = [
+          '%title' => $agency_component->getTitle(),
+          'link' => $agency_component->toLink($this->t('Edit Component'), 'edit-form')->toString(),
+        ];
+      }
+    }
+    // If there isn't an associated Agency Component log an error.
+    else {
+      $error_message = 'Unassociated form: The form, %title, is not associate with an Agency Component.';
```
dangur: @Tom-Camp associate should be associated

docroot/modules/custom/foia_webform/src/Plugin/WebformHandler/FoiaEmailWebformHandler.php
```diff
@@ -0,0 +1,134 @@
...
+    // Get form submissions.
+    $data = $webform_submission->getData();
+    $error_message = NULL;
+    $context = [];
+
+    // If there is a file attachment, get the file URL.
+    if (isset($data['attachments_supporting_documentation'])) {
+      $file = File::load($data['attachments_supporting_documentation']);
+      $file_url = file_create_url($file->getFileUri());
+      $data['attachments_supporting_documentation'] = $file_url;
```
malikkotob: A couple of comments:
1) In #69 I modified this field to be able to accept multiple attachments, so we'll need to modify the code here to handle that.
2) Will this actually 'attach' the file to the e-mail? Or simply include a url to the file? I ask because we're planning on deleting submission data after it is received by the component (another ticket), so a link to the file wouldn't work once that's in place

docroot/modules/custom/foia_webform/src/Plugin/WebformHandler/FoiaEmailWebformHandler.php
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    // Format the submission values as CSV.
+    $submissions = $this->arrayToString($data);
```
malikkotob: Giving this variable a plural structure (submissions vs submission) makes it sound like we're sending multiple submissions. Was there a reason you gave it that name?

docroot/modules/custom/foia_webform/src/Plugin/WebformHandler/FoiaEmailWebformHandler.php
```diff
@@ -0,0 +1,134 @@
...
+    return $csv;
+  }
+
+  /**
+   * Queries for an associated Agency Component given a form ID.
+   *
+   * @param string $form_id
+   *   The form ID.
+   *
+   * @return object
```
malikkotob: I think we can specify here that we're returning the type `NodeInterface` or `null`, e.g.:
```
@return \Drupal\node\NodeInterface|null
```

docroot/modules/custom/foia_webform/src/Plugin/WebformHandler/FoiaEmailWebformHandler.php
```diff
@@ -0,0 +1,134 @@
...
+        $error_message = 'No Submission Email: Unable to send email for %title';
+        $context = [
+          '%title' => $agency_component->getTitle(),
+          'link' => $agency_component->toLink($this->t('Edit Component'), 'edit-form')->toString(),
+        ];
+      }
+    }
+    // If there isn't an associated Agency Component log an error.
+    else {
+      $error_message = 'Unassociated form: The form, %title, is not associated with an Agency Component.';
```
malikkotob: Random thoughts:
1) I'm wondering if this should actually be considered an error. If we ever have a webform on the site that isn't related to the portal, it may not need to be associated with a component.
2) It would be nice if we only added this email handler to webforms associated with components (in case we have webforms that don't need to send out e-mails), but that's not very easy with the current architecture.

Again, these are random thoughts and discussion points, nothing to kick the ticket back for.

docroot/modules/custom/foia_webform/foia_webform.module
```diff
@@ -0,0 +1,55 @@
...
+      'bcc_mail' => '',
+      'bcc_options' => [],
+      'from_mail' => 'default',
+      'from_options' => [],
+      'from_name' => 'default',
+      'subject' => '[webform_submission:values:subject:value]',
+      'body' => '[webform_submission:values:message:value]',
+      'excluded_elements' => [],
+      'html' => TRUE,
+      'attachments' => FALSE,
```
malikkotob: Maybe this should be `TRUE`?

It looks like webform is supposed to handle attachments out of the box: http://cgit.drupalcode.org/webform/tree/src/Plugin/WebformHandler/EmailWebformHandler.php#n780

docroot/modules/custom/foia_webform/src/Plugin/WebformHandler/FoiaEmailWebformHandler.php
Tom-Camp: Good point. I will change it to `$csv_string` for clarity.

PR #64, opened 2017-09-14
Include data diagram in README

adborden:
Not sure where the best place to include this, but I feel like it is helpful to have outside of drive. It seems more relevant to the API side of things rather than the front stage repo.


adborden:
This resolves https://github.com/18F/beta.foia.gov/issues/45

README.md
```diff
@@ -20,3 +20,10 @@ Ansible - 2.2.3
 
 * [Issue queue](https://github.com/18F/beta.foia.gov/issues)
 * [GitHub](https://github.com/usdoj/foia)
+
+### Data model
+
+This diagram illustrates the data model, including what actors are responsible
+for different types of data.
+
+![Diagram describing the data flow of the system](./foia-api-data-diagram.png)
```
malikkotob: Thanks for taking the initiative to add this to the repo! Mind changing the location of the image to live in this [assets directory](https://github.com/usdoj/foia/tree/develop/docs/assets) instead?

adborden: :+1: should I also move this content into `docs/README.md` instead?

brockfanning:
Closing this for now, because this workflow may be changing.

PR #65, opened 2017-09-15
DOJ-94: Tests request submission via email.

dangur:


tests/behat/features/Request.feature
```diff
@@ -0,0 +1,91 @@
...
+    ### Create agency component for testing.
+
+    And I am at 'node/add/agency_component'
+    And for 'Agency Component Name' I enter 'Test agency component'
+
+    ### Associate request form.
+
+    And I select "Test webform" from "Request Submission Form"
+
+    ### Error II (No associated email)
```
ba66e77: This isn't a big enough issue to fail the ticket, but should this failure case test not be a separate test?

composer.json
```diff
@@ -15,7 +15,8 @@
         "drupal/migrate_plus": "^4.0",
         "drupal/migrate_tools": "^4.0",
         "drupal/simplesamlphp_auth": "^3.0",
-        "drupal/jsonapi_extras": "1.x-dev#1a57e11212224129ddb0229726dc7b852bcd836c"
+        "drupal/jsonapi_extras": "1.x-dev#1a57e11212224129ddb0229726dc7b852bcd836c",
+        "drupal/maillog": "1.x-dev#14fd24cd89e299ad271608cd52706c60b753c26d"
```
malikkotob: Is this module just helpful for development/debugging purposes, or do we need it for some functionality? I'm inclined to say let's get rid of this comment unless there's a need for it.

If there is a need, let's commit the `composer.lock` file in an additional commit as well.

docroot/modules/custom/foia_webform/src/Plugin/WebformHandler/FoiaEmailWebformHandler.php
```diff
@@ -101,7 +101,7 @@ public function sendMessage(WebformSubmissionInterface $webform_submission, arra
    *   Returns the array as a comma separated string.
    */
   public function arrayToString(array $data) {
-    $handle = fopen('php://temp', 'c');
+    $handle = fopen('php://temp', 'w');
```
malikkotob: Wondering why we changed this back to `w`. I thought you had looked at the documentation for `fopen` and found that `c` would work better for us?

dangur: In theory <c> would be better, but the test didn't pass, so I changed it back to <w>.

PR #66, opened 2017-09-15
DOJ-119: Migration clean up

Tom-Camp:


docroot/modules/custom/foia_migrate/config/install/migrate_plus.migration.agency_component.yml
```diff
@@ -171,7 +182,7 @@ process:
   field_submission_address/administrative_area: state
   field_submission_address/postal_code: zip
   field_submission_fax: fax
-  field_submission_email: email_submission
+  field_submission_email: email
```
malikkotob: So this is a little tricky because I _think_ there was one agency component where I actually manually put an email address in an `email_submission` key in the source data. We'll have to change that component manually to move that email address to an `emails` key instead, so that it is picked up here.

docroot/modules/custom/foia_migrate/src/data-groomer.php
```diff
@@ -103,6 +103,20 @@ function extract_personnel_from_components(array &$components) {
...
+      foreach ($misc_users as $title => $miscellaneous) {
+        $misc = set_personnel_id($foia_personnel, $component, 'miscellaneous', $id, $miscellaneous);
+        if ($misc) {
+          assign_agency_name_to_personnel($component, $misc);
+          $misc->title = $title;
+          $foia_personnel['personnel'][] = $misc;
+        }
+      }
+      unset($component->misc);
+    }
```
malikkotob: looks good!

docroot/modules/custom/foia_migrate/src/data-groomer.php
```diff
@@ -130,25 +144,42 @@ function set_component_id(&$component, &$id) {
...
+    $component->{$personnel_type} = new stdClass();
+  }
   $existing_id = '';
   if (isset($personnel['personnel'])) {
     $existing_id = get_existing_personnel_id($personnel['personnel'], $component, $individual_personnel);
   }
   if ($existing_id) {
-    $component->{$personnel_type}->id = $existing_id;
+    if ($personnel_type === 'miscellaneous') {
+      $component->miscellaneous->{$existing_id} = $existing_id;
```
malikkotob: Definitely not a reason to hold back the PR, but would the script work if we modified this to be:
```
$component->miscellaneous->id[] = $existing_id;
```
Wondering if we can keep the data looking consistent. Right now we have:
```
            "service_center": {
                "id": 3
            },
            "miscellaneous": {
                "4": 4
            }
```
vs something like this:
```
            "service_center": {
                "id": 3
            },
            "miscellaneous": {
                "id": [
                  4
                ]
            }
```

docroot/modules/custom/foia_migrate/src/data-groomer.php
```diff
@@ -130,25 +144,42 @@ function set_component_id(&$component, &$id) {
...
+    }
+    else {
+      $component->{$personnel_type}->id = $existing_id;
+    }
     return FALSE;
   }
   else {
-    $component->{$personnel_type}->id = $id;
+    if ($personnel_type === 'miscellaneous') {
+      $component->miscellaneous->{$id} = $id;
```
malikkotob: Same comment here as [above](https://github.com/usdoj/foia/pull/66/files#r140275813)

malikkotob:
This needs a rebase and a config export from the latest stage DB

PR #67, opened 2017-09-15
DOJ-97: Agency users can update any component within their agency 

Tom-Camp:


docroot/modules/custom/foia_users/foia_users.module
```diff
@@ -7,30 +7,25 @@
 
 use Drupal\Core\Access\AccessResult;
 use Drupal\user\Entity\User;
+use Drupal\Core\Entity\EntityInterface;
+use Drupal\Core\Session\AccountInterface;
 
 /**
  * Validates Office NID matching for users.
  */
 function _office_user_nid_access($node, $account) {
```
malikkotob: We should change the name of this function to be `_agency_user_nid_access`

docroot/modules/custom/foia_users/foia_users.module
```diff
@@ -7,30 +7,25 @@
...
+  if ($user_agency === $node_agency) {
     $result = AccessResult::allowed()
       ->cachePerUser()
       ->addCacheableDependency($node);
   }
-  else {
-    $result = AccessResult::forbidden()
-      ->cachePerUser()
-      ->addCacheableDependency($node);
-  }
```
malikkotob: Why did we remove this snippet?

docroot/modules/custom/foia_users/foia_users.module
```diff
@@ -40,35 +35,57 @@ function _office_user_nid_access($node, $account) {
...
+    // Get the current user's roles.
     $user = User::load(\Drupal::currentUser()->id());
     $user_roles = $user->getRoles();
 
-    if (in_array("foia_office_creator", $user_roles) || in_array("foia_office_reviewer", $user_roles)) {
+    if (in_array("agency_manager", $user_roles)) {
+      switch ($op) {
+        case 'update':
+          // Check if the user has the proper Agency association.
+          return _office_user_nid_access($node, $account);
```
malikkotob: Same comment as above, we should rename this function to `_agency_user_nid_access`

docroot/modules/custom/foia_users/foia_users.module
```diff
@@ -77,13 +94,12 @@ function foia_users_node_access($node, $op, $account) {
  * Custom submit handler for office value.
  */
 function _form_validation_foiaoffice($form, $form_state) {
```
malikkotob: We should change the comments above to reference 'agency' instead of 'office', we should also change the function name.

docroot/modules/custom/foia_users/foia_users.module
```diff
@@ -77,13 +94,12 @@ function foia_users_node_access($node, $op, $account) {
...
-
-  if ($form_state->hasValue('field_foia_office')) {
+  if ($form_state->hasValue('field_agency')) {
 
     $role_array = ["foia_office_creator", "foia_office_reviewer"];
     $user_role = $form_state->getValue('roles');
     $user_role_match = FALSE;
-    $field_foia_office = $form_state->getValue('field_foia_office')[0]['target_id'];
+    $field_foia_office = $form_state->getValue('field_agency')[0]['target_id'];
```
malikkotob: this should be renamed to `$user_agency` or `$field_foia_agency`

docroot/modules/custom/foia_users/foia_users.module
```diff
@@ -92,11 +108,10 @@ function _form_validation_foiaoffice($form, $form_state) {
     }
 
     if (is_null($field_foia_office) && ($user_role_match)) {
-      $form['field_foia_office']['widget']['#required'] = TRUE;
-      $form_state->setErrorByName('field_foia_office', t('The roles "FOIA Office creator" and "FOIA Office reviewer" must be assigned to an office.'));
+      $form['field_agency']['widget']['#required'] = TRUE;
+      $form_state->setErrorByName('field_agency', t('The roles "Agency Component creator" and "Agency Component reviewer" must be assigned to an office.'));
```
malikkotob: So is the code here essentially attempting to force someone who's editing this user to fill out the agency field if they have certain roles? If so, we should probably be making sure they have the new 'Agency Manager' role, and not these other roles.

config/default/core.entity_form_display.user.user.default.yml
```diff
@@ -3,7 +3,7 @@ langcode: en
 status: true
 dependencies:
   config:
-    - field.field.user.user.field_foia_office
+    - field.field.user.user.field_agency
```
malikkotob: We need to prevent users from being able to edit their own agency on their /user/uid/edit pages

malikkotob:
Additional feedback:
1) We need to adjust the permissions for the Agency Manager role for some of the implicit items permissions that the role will need. I think these are only:
-Create new FOIA Personnel entities
-Delete FOIA Personnel entities
-Edit FOIA Personnel entities
2) We should prevent the Agency Manager (and really all users) from being able to change the agency their components are associated with. I *think* core supports this out of the box using custom permissions on the field and only letting the Admin have edit rights. Others should be allowed to View, but not edit.
image

docroot/modules/custom/foia_users/foia_users.module
Tom-Camp: Interesting. I simply modified code that was already there to work with `field_agency` rather than `field_foia_office`.

docroot/modules/custom/foia_users/foia_users.module
```diff
@@ -40,63 +40,83 @@ function _office_user_nid_access($node, $account) {
...
-function _form_validation_foiaoffice($form, $form_state) {
-
-  if ($form_state->hasValue('field_foia_office')) {
-
-    $role_array = ["foia_office_creator", "foia_office_reviewer"];
+function _form_validation_foia_agency($form, $form_state) {
+  if ($form_state->hasValue('field_agency')) {
+    $role_array = [
+      'foia_office_creator',
+      'foia_office_reviewer',
```
malikkotob: We should remove these roles since they are obsolete.

docroot/modules/custom/foia_users/foia_users.module
```diff
@@ -40,63 +40,83 @@ function _office_user_nid_access($node, $account) {
...
+    if (is_array($user_role)) {
+      $user_role_match = array_intersect($user_role, $role_array) ? TRUE : $user_role_match;
     }
 
-    if (is_null($field_foia_office) && ($user_role_match)) {
-      $form['field_foia_office']['widget']['#required'] = TRUE;
-      $form_state->setErrorByName('field_foia_office', t('The roles "FOIA Office creator" and "FOIA Office reviewer" must be assigned to an office.'));
+    if (is_null($user_agency) && ($user_role_match)) {
+      $form['field_agency']['widget']['#required'] = TRUE;
+      $form_state->setErrorByName('field_agency', t('The roles "Agency Component creator", "Agency Component reviewer" and "Agency Manager" must be assigned to an agency.'));
```
malikkotob: We should get rid of mention of `Agency Component creator` and `Agency Component reviewer` from this message since I don't think we'll be using it. Also, maybe we can say 

> Users with the "Agency Manager" role must be assigned to an agency.

docroot/modules/custom/foia_users/foia_users.module
```diff
@@ -40,63 +40,83 @@ function _office_user_nid_access($node, $account) {
...
- * Custom submit handler for office value.
+ * Custom submit handler for agency value.
  */
-function _form_validation_foiaoffice($form, $form_state) {
-
-  if ($form_state->hasValue('field_foia_office')) {
-
-    $role_array = ["foia_office_creator", "foia_office_reviewer"];
+function _form_validation_foia_agency($form, $form_state) {
+  if ($form_state->hasValue('field_agency')) {
```
malikkotob: If a condition to showing the error message below is that the form has a value for `field_agency`, it will never display, will it?

Tom-Camp: `field_agency` is the entity reference field to the Agency taxonomy term. The `hasValue` evaluates to true if that field exists. It works more like `array_key_exists()`.

docroot/modules/custom/foia_users/foia_users.module
Tom-Camp: Agreed. Should I not be checking for those other roles at all? That would simplify the `array_intersect` to an `in_array`.

malikkotob: I don't think so - see [comment above](https://github.com/usdoj/foia-api/pull/67#discussion_r141098128)

docroot/modules/custom/foia_users/foia_users.module
malikkotob: You're right! this worked.

malikkotob:
@Tom-Camp I realized I left off "View published FOIA Personnel entities" as a permission that Agency Managers should also have. When viewing their agency component they aren't seeing the personnel entities currently

config/default/core.entity_form_display.user.user.default.yml
malikkotob: This is working now!

config/default/user.role.authenticated.yml
```diff
@@ -12,4 +12,6 @@ permissions:
   - 'access content'
   - 'access shortcuts'
   - 'access site-wide contact form'
+  - 'view field_agency'
   - 'view media'
+  - 'view own field_agency'
```
malikkotob: Do you know what these are/why they were added here?

Tom-Camp: Well, I guess we could get rid of that as long as they have permission to `View anyone's value for field field_agency`

PR #68, opened 2017-09-15
DOJ-0: Fixed overridden configuration.

danepowell:
Some configuration has not been exported correctly, causing it to be overridden whenever you do a `config-import`. Ideally after you run `config-import` once, subsequent calls should have nothing to import. If they do it's a sign of an incomplete or corrupt export.

### DIFF
--- a/config/default/core.entity_form_display.scheduled_update.multiple_node_embargo.default.yml
+++ b/config/default/core.entity_form_display.scheduled_update.multiple_node_embargo.default.yml
@@ -18,7 +18,7 @@ content:
     region: content
     settings:
       match_operator: CONTAINS
-      size: '60'
+      size: 60
       placeholder: ''
     third_party_settings: {  }
   field_moderation_state:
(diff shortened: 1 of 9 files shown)


malikkotob:
@danepowell, I'm wondering if this is being caused by a difference in `drush` versions? I'm not seeing these items.

danepowell:
Unlikely to change based on Drush version. If you install the site from scratch and try to import the configuration, you should notice it constantly trying to import these config files.

malikkotob:
Interesting, okay I'll try to reproduce.

danepowell:
Here's how I reproduced (starting from scratch):
1. Clone repo
2. `composer install`
3. `vagrant up`
4. `blt setup` (this will fail, so run the next step. we should figure out why)
5. `drush @dojfoia.local cron` && `drush @dojfoia.local cr` (to fix the broken install)
6. `drush @dojfoia.local cim`

At this point, the active configuration (i.e. in the database) should match config on disk, but it does not. If you run `drush @dojfoia.local cex`, it will dump the same files you see in this PR, showing that the configuration on disk is corrupt/stale and needs to updated.

PR #73, opened 2017-09-26
DOJ-128: CORS config for local/dev/stage.

adborden:
This adds `http://localhost:4000` as an acceptable origin for API requests. This should only be enabled for local/dev/staging, so let me know if there is a better config to add this.

Prod will have a config for `www.foia.gov` or `beta.foia.gov`, but I'm not sure where to put that yet.


adborden:
@malikkotob pointed out that the API could be used from JS in a browser, so we want to allow all origins in prod and the other environments as well.

adborden:
We should limit the CORS config to just the API, since the admin endpoints should still only be reachable by the front end.

malikkotob:
This has been merged into `develop`.

PR #76, opened 2017-09-28
DOJ-140: Creating Agency Administrator role

Tom-Camp:


config/default/user.role.agency_administrator.yml
```diff
@@ -0,0 +1,43 @@
...
+  - 'edit own agency_component content'
+  - 'edit terms in agency'
+  - 'revert agency_component revisions'
+  - 'revert all foia personnel revisions'
+  - 'use text format rich_text'
+  - 'view agency_component revisions'
+  - 'view all foia personnel revisions'
+  - 'view published foia personnel entities'
+  - 'view the administration theme'
+  - 'view unpublished foia personnel entities'
```
malikkotob: Some permissions that should be granted/removed for the administrator:
- Grant `Edit own value for field field_agency` (they're allowed to edit others' values)
- Grant `View own value for field field_agency` (they're allowed to view others' values)
- Grant `View own unpublished content` (since they have the ability to create components)
- Grant `Access the webform overview page` (since these are the users, like Matt, who may need to create/tweak webforms for components)
- Grant `Create webforms` (see above reason)
- Grant `Delete any webform`
- Grant `Delete own webform`
- Grant `Edit any webform`
- Grant `Edit own webform`
- Grant `View any webform submission`
- Grant `View own webform submission`
- Remove `Agency Component: Delete revisions` (wouldn't want audit trails to be wiped away)
- Remove `Delete all revisions` for FOIA personnel entities (same audit trails reason)

config/default/user.role.agency_administrator.yml
```diff
@@ -0,0 +1,43 @@
...
+third_party_settings:
+  lightning_core:
+    description: ''
+id: agency_administrator
+label: 'Agency Administrator'
+weight: 11
+is_admin: null
+permissions:
+  - 'access content overview'
+  - 'access devel information'
```
malikkotob: I don't think we would want them to have access to `devel` info. Rather than removing this, though, maybe you can just disable `devel` as part of this ticket.

composer.json
```diff
@@ -17,7 +17,8 @@
         "drupal/simplesamlphp_auth": "^3.0",
         "drupal/jsonapi_extras": "1.x-dev#1a57e11212224129ddb0229726dc7b852bcd836c",
         "drupal/maillog": "1.x-dev#14fd24cd89e299ad271608cd52706c60b753c26d",
-        "drupal/restui": "^1.15"
+        "drupal/restui": "^1.15",
+        "drupal/roleassign": "^1.0@alpha"
```
malikkotob: Just a tip! Splitting up changes to `composer.json` and `composer.lock` into separate commits makes certain things easier, like rebasing for example. If you were to rebase onto the latest `develop` branch and it happened to have `composer.lock` changes, you could basically do away with your separate `composer.lock` commit and just run `composer install` again with your `composer.json` commit still intact. We used to do something similar back in D7 with make files, where we'd have one commit to update the file and a separate to run it. Similar concept with sass changes in one commit + compiled css changes in a separate one.

docroot/modules/custom/foia_users/foia_users.module
```diff
@@ -122,4 +178,25 @@ function foia_users_form_alter(&$form, $form_state, $form_id) {
   if ($form_id == ('user_form' || 'user_register_form')) {
     $form['#validate'][] = '_form_validation_foia_agency';
   }
+
+  if ($form_id === 'user_register_form') {
+    $current_user = User::load(\Drupal::currentUser()->id());
+    $current_user_roles = $current_user->getRoles();
+    // Check the Agency Manager checkbox.
+    if (in_array('agency_administrator', $current_user_roles)) {
+      $form['account']['roles']['#default_value'][] = 'agency_manager';
```
malikkotob: Discussion point and not MVP: I'm wondering if we should remove this from the agency admin's own `/user/uid/edit` page. A bit confusing I think.
image

config/default/user.role.agency_administrator.yml
dangur: I'm not sure the field_agency changes are necessary. It looks like `Agency Administrator`s can already edit and view that value in the `Agency Component` node edit form.

malikkotob:
@Tom-Camp, fix-ups look great. Mind just disabling the `devel` module and rebasing against the latest `develop` branch? In the mean time, I'm going to go ahead and deploy this to the dev environment for UAT.

PR #78, opened 2017-10-02
DOJ-120: Adding form queue handler.

Tom-Camp:


docroot/modules/custom/foia_webform/src/Plugin/WebformHandler/FoiaSubmissionQueueHandler.php
```diff
@@ -0,0 +1,42 @@
...
+ *   results = \Drupal\webform\WebformHandlerInterface::RESULTS_PROCESSED,
+ * )
+ */
+class FoiaSubmissionQueueHandler extends EmailWebformHandler {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function sendMessage(WebformSubmissionInterface $webform_submission, array $message) {
+    // @var @var QueueFactory $queue_factory
```
malikkotob: is `@var` in here twice on purpose?

docroot/modules/custom/foia_webform/src/Plugin/WebformHandler/FoiaSubmissionQueueHandler.php
```diff
@@ -0,0 +1,42 @@
...
+    $queue_factory = \Drupal::service('queue');
+
+    // @var QueueInterface $queue
+    $queue = $queue_factory->get('foia_queue_submissions');
```
malikkotob: Maybe call this something like `foia_submissions_queue` or just `foia_submissions`. The current name sounds kind of like an action.

docroot/modules/custom/foia_webform/src/Plugin/WebformHandler/FoiaSubmissionQueueHandler.php
```diff
@@ -0,0 +1,42 @@
...
+    $submission = new \stdClass();
+    $submission->sid = $webform_submission->id();;
+    $submission->message = $message;
+
+    // Log the form submission.
+    \Drupal::logger('foia_webform')
+      ->info('FOIA form submission added to queue.', ['link' => $webform_submission->toLink($this->t('View'))->toString()]);
```
malikkotob: Good thought to log this. Slipped my mind. Mind adding the submission ID to the log message?

docroot/modules/custom/foia_webform/src/Plugin/QueueWorker/FoiaFormQueue.php
```diff
@@ -0,0 +1,38 @@
...
+
+use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
+use Drupal\Core\Queue\QueueWorkerBase;
+use Drupal\webform\WebformSubmissionInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Provides base functionality for the FoiaForm Queue Workers.
+ */
+abstract class FoiaFormQueue extends QueueWorkerBase implements ContainerFactoryPluginInterface {
```
malikkotob: Maybe we should give the worker a similarly formatted name? e.g. `FoiaSubmissionQueueWorker`

docroot/modules/custom/foia_webform/src/Plugin/WebformHandler/FoiaSubmissionQueueHandler.php
Tom-Camp: No

docroot/modules/custom/foia_webform/src/Plugin/WebformHandler/FoiaSubmissionQueueHandler.php
```diff
@@ -0,0 +1,65 @@
...
+   *   The webform submission object.
+   * @param array $message
+   *   The email message array.
+   */
+  private function queueSubmission(WebformSubmissionInterface $webform_submission, array $message) {
+    // @var QueueFactory $queue_factory
+    $queue_factory = \Drupal::service('queue');
+
+    // @var QueueInterface $queue
+    $foia_submission_queue = $queue_factory->get('foia_form_manual_submitter');
```
malikkotob: I'm not sure that a queue can be described as a `submitter`. We might call the worker class a submitter if its submitting the items that it's processing, but the queue itself should probably just be named to describe its contents (e.g. `submissions`, `foia_submissions` etc.).

docroot/modules/custom/foia_webform/src/Plugin/WebformHandler/FoiaSubmissionQueueHandler.php
```diff
@@ -0,0 +1,65 @@
...
+    $submission = new \stdClass();
+    $submission->sid = $webform_submission->id();;
```
malikkotob: extra semicolon here

docroot/modules/custom/foia_webform/src/Plugin/WebformHandler/FoiaSubmissionQueueHandler.php
```diff
@@ -0,0 +1,65 @@
...
+    $submission->message = $message;
+
+    // Log the form submission.
+    \Drupal::logger('foia_webform')
+      ->info('FOIA form submission #%sid added to queue.',
+        [
+          '%sid' => $webform_submission->id(),
```
malikkotob: nice!

docroot/modules/custom/foia_webform/src/Plugin/QueueWorker/FoiaSubmissionQueueWorker.php
```diff
@@ -0,0 +1,38 @@
...
+
+use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
+use Drupal\Core\Queue\QueueWorkerBase;
+use Drupal\webform\WebformSubmissionInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Provides base functionality for the FoiaForm Queue Workers.
+ */
+abstract class FoiaSubmissionQueueWorker extends QueueWorkerBase implements ContainerFactoryPluginInterface {
```
malikkotob: Because we're only planning on having a single queue worker (we don't plan on leveraging Drupal's built in Cron system, so we won't have a manual vs cron worker) I'm not sure there is added value in implementing this as an abstract class. I'm thinking we can just have this as our concrete worker (i.e. let's add the `@QueueWorker` annotation here)

docroot/modules/custom/foia_webform/src/Plugin/QueueWorker/FoiaSubmissionQueueWorker.php
```diff
@@ -0,0 +1,38 @@
...
+
+  /**
+   * The form submission.
+   *
+   * @var \Drupal\webform\WebformSubmissionInterface
+   */
+  protected $formSubmission;
```
malikkotob: Is this not an instance of  `EntityStorageInterface`?

docroot/modules/custom/foia_webform/src/Plugin/QueueWorker/FoiaSubmissionQueueWorker.php
```diff
@@ -0,0 +1,38 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public function __construct(WebformSubmissionInterface $submission) {
```
malikkotob: Same comment here - think this should be `WebformSubmissionStorageInterface $webformSubmissionStorage`

docroot/modules/custom/foia_webform/src/Plugin/WebformHandler/FoiaSubmissionQueueHandler.php
```diff
@@ -0,0 +1,65 @@
...
+    if (in_array($state, $this->configuration['states'])) {
+      // Get the Webform generated email message.
+      $message = $this->getMessage($webform_submission);
+      // Add the submission to the queue.
+      $this->queueSubmission($webform_submission, $message);
+    }
+  }
+
+  /**
+   * Adds the submission to the foia_form_manual_submitter queue.
```
malikkotob: This comment needs to be cleaned up to reference the right queue name. Or you can keep it generic and say that the submission is added to a queue - whatever your prefer!

docroot/modules/custom/foia_webform/src/Plugin/QueueWorker/FoiaSubmissionQueueWorker.php
```diff
@@ -0,0 +1,15 @@
...
+
+/**
+ * Provides functionality for working with the queued FOIA form submissions.
+ *
+ * @QueueWorker (
+ *   id = "foia_submissions",
+ *   title = @Translation("Manual FOIA Form submitter"),
+ * )
+ */
+class FoiaSubmissionQueueWorker extends QueueWorker {}
```
malikkotob: you had this right previously - must've been a typo, but this should extend `QueueWorkerBase`

malikkotob:
Covered by #79, which was merged.

PR #80, opened 2017-10-03
DOJ-120: Get tests passing temporarily.

dangur:


malikkotob:
Merged into `develop`

PR #83, opened 2017-10-05
DOJ-121: Send out request submissions to agency components via API

Tom-Camp:
@malikkotob - this is pretty rough. It needs some work and I'll clean up comments in the a.m.


docroot/modules/custom/foia_webform/src/AgencyLookupService.php
```diff
@@ -0,0 +1,34 @@
+<?php
+
+namespace Drupal\foia_webform;
+
+use Drupal\node\Entity\Node;
+
+/**
+ * Class AgencyLookupService.
+ *
+ * @package Drupal\foia_webform
```
malikkotob: Don't think you need the `@package` annotation

docroot/modules/custom/foia_webform/src/AgencyLookupService.php
```diff
@@ -0,0 +1,34 @@
...
+
+  /**
+   * Queries for an associated Agency Component given a form ID.
+   *
+   * @param string $webformId
+   *   The form ID.
+   *
+   * @return \Drupal\node\NodeInterface|null
+   *   The Agency Component object or NULL.
+   */
```
malikkotob: Since this class implements an interface, we can just use the `inheritdoc` notation for the doc block for any methods that have doc blocks defined on the interface:
```
  /**
   * {inheritdoc}
   */
```

docroot/modules/custom/foia_webform/src/AgencyLookupService.php
```diff
@@ -0,0 +1,34 @@
...
+  public function getComponentByWebform($webformId) {
+    $entity_query_service = \Drupal::service('entity.query');
+    $query = $entity_query_service->get('node')
```
malikkotob: Let's use camel casing throughout to establish a team convention

docroot/modules/custom/foia_webform/src/AgencyLookupServiceInterface.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+namespace Drupal\foia_webform;
+
+/**
+ * Interface AgencyLookupServiceInterface.
```
malikkotob: This should be some sort of minimal description about the, rather than stating that it's an interface. Something like: `Provides an interface defining an agency lookup service.` You can check out modules in core for other examples.

docroot/modules/custom/foia_webform/src/AgencyLookupServiceInterface.php
```diff
@@ -0,0 +1,21 @@
...
+  /**
+   * Looks up the Agency Component based on the associated webform.
+   *
+   * @param string $webformId
+   *   The form ID.
+   *
+   * @return \Drupal\node\NodeInterface|null
+   *   The Agency Component object or NULL.
+   */
+  public function getComponentByWebform($webformId);
```
malikkotob: Nice!

docroot/modules/custom/foia_webform/src/FoiaSubmissionApiService.php
```diff
@@ -0,0 +1,95 @@
...
+
+use Drupal\foia_webform\AgencyLookupServiceInterface;
+use Drupal\webform\WebformSubmissionInterface;
+use GuzzleHttp\ClientInterface;
+use Drupal\node\Entity\Node;
+
+/**
+ * Class FoiaSubmissionApiService.
+ *
+ * @package Drupal\foia_webform\Plugin\QueueWorker
```
malikkotob: No need for `@package` annotation

docroot/modules/custom/foia_webform/src/FoiaSubmissionApiService.php
```diff
@@ -0,0 +1,95 @@
...
+ */
+class FoiaSubmissionApiService implements FoiaSubmissionServiceInterface {
+
+  private $apiURL;
```
malikkotob: Let's omit this as a property on the class. Since we're only using it within the same method that we set it in, we don't buy much by having it as a property.

docroot/modules/custom/foia_webform/src/FoiaSubmissionApiService.php
```diff
@@ -0,0 +1,95 @@
...
+
+  /**
+   * The Agency Lookup service.
+   *
+   * @var \Drupal\foia_webform\AgencyLookupService
+   */
+  protected $agencyLookUpService;
+
+  /**
+   * The HTTP client to make calls to the API with.
```
malikkotob: In our case, we're not using this to make calls to an API. Rather, we're using it send data off to endpoints.

docroot/modules/custom/foia_webform/src/FoiaSubmissionApiService.php
```diff
@@ -0,0 +1,95 @@
...
+   * @param \Drupal\foia_webform\AgencyLookupServiceInterface $agencyLookupService
+   *   The Agency Lookup service.
+   */
+  public function __construct(ClientInterface $httpClient, AgencyLookupServiceInterface $agencyLookupService) {
+    $this->httpClient = $httpClient;
+    $this->agencyLookUpService = $agencyLookupService;
+  }
+
+  /**
+   * Submits the submission via the API.
```
malikkotob: Maybe instead: `Submits the submission to a component's endpoint.`

docroot/modules/custom/foia_webform/src/FoiaSubmissionApiService.php
```diff
@@ -0,0 +1,95 @@
...
+   *
+   * @param \Drupal\webform\WebformSubmissionInterface $webformSubmission
+   *   The webform submission.
+   * @param \Drupal\node\Entity\Node $agencyComponent
+   *   The Agency Components node object.
+   */
+  public function sendSubmissionToComponent(WebformSubmissionInterface $webformSubmission, Node $agencyComponent) {
```
malikkotob: The same way we used `WebformSubmissionInterface`, let's use `NodeInterface` here instead.

docroot/modules/custom/foia_webform/src/FoiaSubmissionApiService.php
```diff
@@ -0,0 +1,95 @@
...
+    $this->httpClient = $httpClient;
+    $this->agencyLookUpService = $agencyLookupService;
+  }
+
+  /**
+   * Submits the submission via the API.
+   *
+   * @param \Drupal\webform\WebformSubmissionInterface $webformSubmission
+   *   The webform submission.
+   * @param \Drupal\node\Entity\Node $agencyComponent
```
malikkotob: The same way we used `WebformSubmissionInterface`, let's use `NodeInterface` here instead.

docroot/modules/custom/foia_webform/src/FoiaSubmissionApiService.php
```diff
@@ -0,0 +1,95 @@
...
+   *   The Agency Components node object.
+   */
+  public function sendSubmissionToComponent(WebformSubmissionInterface $webformSubmission, Node $agencyComponent) {
+    $this->apiURL = $agencyComponent->get('field_submission_api');
```
malikkotob: If we remove `apiUrl` as a property on the class, this will just become `$apiUrl` instead of `$this->apiUrl`

docroot/modules/custom/foia_webform/src/FoiaSubmissionApiService.php
```diff
@@ -0,0 +1,95 @@
...
+
+    if ($this->apiURL) {
```
malikkotob: This would also become `$apiUrl`

docroot/modules/custom/foia_webform/src/FoiaSubmissionApiService.php
```diff
@@ -0,0 +1,95 @@
...
+    return $webformSubmission->getData();
+  }
+
+  /**
+   * Parse the Agency information from the Agency Component.
+   *
+   * @param \Drupal\node\Entity\Node $agencyComponent
+   *   The Agency Component node object.
+   */
+  private function getAgencyInfo(Node $agencyComponent) {
```
malikkotob: This actually seems like it might be a good candidate for a method to add to the `AgencyLookupService`.

docroot/modules/custom/foia_webform/src/FoiaSubmissionApiService.php
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  /**
+   * Parse the submission values into an array.
+   *
+   * @param \Drupal\webform\WebformSubmissionInterface $webformSubmission
+   *   The Webform submission values.
+   *
+   * @return array
+   *   Returns the submissions as an array.
```
malikkotob: `submission` instead of `submissions`

docroot/modules/custom/foia_webform/src/Plugin/QueueWorker/FoiaSubmissionQueueWorker.php
```diff
@@ -19,29 +22,76 @@ class FoiaSubmissionQueueWorker extends QueueWorkerBase implements ContainerFact
...
   }
 
   /**
    * {@inheritdoc}
    */
-  public function processItem($data) {}
+  public function processItem($data) {
+    $webformSubmission = $this->webformStorage->load($data->sid);
+
+    $agencyComponent = $this->agencyLookUpService->getComponentByWebform($webformSubmission->getWebform()->getOriginalId());
```
malikkotob: use `id` instead of `originalId`

docroot/modules/custom/foia_webform/src/FoiaSubmissionApiService.php
```diff
@@ -0,0 +1,95 @@
...
+   *   The webform submission.
+   * @param \Drupal\node\Entity\Node $agencyComponent
+   *   The Agency Components node object.
+   */
+  public function sendSubmissionToComponent(WebformSubmissionInterface $webformSubmission, Node $agencyComponent) {
+    $this->apiURL = $agencyComponent->get('field_submission_api');
+
+    if ($this->apiURL) {
+      $payload = $this->getSubmissionPayload($webformSubmission);
+      $agencyInfo = $this->getAgencyInfo($agencyComponent);
```
malikkotob: Is the intent behind the `getSubmissionPayload` method to return the array in the format as defined by the API specification? If so, I would imagine that we'd move the call to `getAgencyInfo` to be within that method so that it can return a payload that's ready to send off.

docroot/modules/custom/foia_webform/src/FoiaSubmissionApiService.php
```diff
@@ -0,0 +1,95 @@
...
+    }
+    else {
+      \Drupal::logger('foia_webform')
+        ->error('Unable to submit request via the API. Missing API Submission URL for %component',
+          [
+            '%component' => $agencyComponent->label(),
```
malikkotob: Unfortunately, some components have the same name across different agencies, so if we want to use their name, we have to include the agency name as well to uniquely identify which component we were submitting to. We can probably get away with just using the component's `nid`.

docroot/modules/custom/foia_webform/src/FoiaSubmissionApiService.php
```diff
@@ -0,0 +1,95 @@
...
+   */
+  public function sendSubmissionToComponent(WebformSubmissionInterface $webformSubmission, Node $agencyComponent) {
+    $this->apiURL = $agencyComponent->get('field_submission_api');
+
+    if ($this->apiURL) {
+      $payload = $this->getSubmissionPayload($webformSubmission);
+      $agencyInfo = $this->getAgencyInfo($agencyComponent);
+    }
+    else {
+      \Drupal::logger('foia_webform')
```
malikkotob: Maybe we can use the pattern we see [here](https://github.com/WhiteHouse/fb_messenger_bot/blob/master/fb_messenger_bot.services.yml#L4) with the logger channel, which is defined [here](https://github.com/WhiteHouse/fb_messenger_bot/blob/master/fb_messenger_bot.services.yml#L15). Let me know if you need help setting that up.

docroot/modules/custom/foia_webform/src/FoiaSubmissionEmailService.php
```diff
@@ -0,0 +1,45 @@
...
+use Drupal\foia_webform\AgencyLookupServiceInterface;
+use Drupal\webform\WebformSubmissionInterface;
+use Drupal\node\Entity\Node;
+
+/**
+ * Class FoiaSubmissionEmailService.
+ *
+ * @package Drupal\foia_webform\Plugin\QueueWorker
+ */
+class FoiaSubmissionEmailService implements FoiaSubmissionServiceInterface {
```
malikkotob: No need to include this code with this PR. Let's keep the PR as relevant to issue 121 as possible.

docroot/modules/custom/foia_webform/src/FoiaSubmissionProcessingFactory.php
```diff
@@ -0,0 +1,49 @@
+<?php
+
+namespace Drupal\foia_webform\Plugin\QueueWorker;
+
+use Drupal\node\Entity\Node;
+
+/**
+ * Class FoiaSubmissionProcessingFactory.
+ *
+ * @package Drupal\foia_webform\Plugin\QueueWorker
```
malikkotob: No need for the `@package` annotation

docroot/modules/custom/foia_webform/src/FoiaSubmissionProcessingFactory.php
```diff
@@ -0,0 +1,49 @@
...
+ */
+class FoiaSubmissionProcessingFactory {
```
malikkotob: Should be `FoiaSubmissionServiceFactory`

docroot/modules/custom/foia_webform/src/FoiaSubmissionProcessingFactory.php
```diff
@@ -0,0 +1,49 @@
...
+
+  /**
+   * Check the submission format and return the service.
+   *
+   * @param \Drupal\node\Entity\Node $agencyComponent
```
malikkotob: let's use `NodeInterface` here

docroot/modules/custom/foia_webform/src/FoiaSubmissionProcessingFactory.php
```diff
@@ -0,0 +1,49 @@
...
+   *   The Agency Component node object.
+   *
+   * @return \Drupal\foia_webform\FoiaSubmissionServiceInterface
+   *   Returns a Fo
+   */
+  public function get(Node $agencyComponent) {
```
malikkotob: same as above, let's use `NodeInterface` instead of `Node`

docroot/modules/custom/foia_webform/src/FoiaSubmissionProcessingFactory.php
```diff
@@ -0,0 +1,49 @@
...
+    $submissionPreference = $agencyComponent->get('field_portal_submission_format');
+
+    switch ($submissionPreference) {
+      case 'api':
+        $serviceName = 'foia_webform.foia_submission_api_service';
+        break;
+
+      case 'email':
```
malikkotob: Let's leave off email-related features for now. We can add it in with 122.

docroot/modules/custom/foia_webform/src/FoiaSubmissionProcessingFactory.php
```diff
@@ -0,0 +1,49 @@
...
+        $serviceName = 'foia_webform.foia_submission_email_service';
+        break;
+
+      default:
+        $serviceName = 'foia_webform.foia_submission_email_service';
```
malikkotob: Same comment as above, let's leave off email-related features for now. We can add it in with 122.

docroot/modules/custom/foia_webform/src/FoiaSubmissionProcessingFactory.php
```diff
@@ -0,0 +1,49 @@
...
+        \Drupal::logger('foia_webform')
```
malikkotob: See previous comment about setting up a logger channel in `services.yml`

docroot/modules/custom/foia_webform/src/FoiaSubmissionProcessingFactory.php
```diff
@@ -0,0 +1,49 @@
...
+          ->notice('Invalid submission preference for component %title, defaulting to email.',
+            [
+              '%title' => $agencyComponent->label(),
```
malikkotob: See previous comment about using the component's name as an identifier.

docroot/modules/custom/foia_webform/src/FoiaSubmissionProcessingFactory.php
```diff
@@ -0,0 +1,49 @@
...
+              'link' => $agencyComponent->toLink($this->t('Edit Component'), 'edit-form')->toString(),
+            ]
+          );
+    }
+
+    return \Drupal::service($serviceName);
```
malikkotob: nice!

docroot/modules/custom/foia_webform/src/FoiaSubmissionProcessingFactoryInterface.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+namespace Drupal\foia_webform;
+
+use Drupal\node\Entity\Node;
+
+/**
+ * Provides an interface for submitting the request using the proper method.
```
malikkotob: maybe instead of `using the proper method` we can say `using the component's preferred method.`

docroot/modules/custom/foia_webform/src/FoiaSubmissionProcessingFactoryInterface.php
```diff
@@ -0,0 +1,20 @@
...
+ */
+interface FoiaSubmissionProcessingFactoryInterface {
```
malikkotob: name should be `FoiaSubmissionServiceFactoryInterface`

docroot/modules/custom/foia_webform/src/FoiaSubmissionProcessingFactoryInterface.php
```diff
@@ -0,0 +1,20 @@
...
+
+  /**
+   * Determines which method to use for submitting the request.
+   *
+   * @param \Drupal\node\Entity\Node $agencyComponent
```
malikkotob: `NodeInterface` instead of `Node`

docroot/modules/custom/foia_webform/src/FoiaSubmissionProcessingFactoryInterface.php
```diff
@@ -0,0 +1,20 @@
...
+   *   The Agency Component node object.
+   */
+  public function get(Node $agencyComponent);
```
malikkotob: `NodeInterface` instead of `Node`

docroot/modules/custom/foia_webform/src/FoiaSubmissionServiceInterface.php
```diff
@@ -0,0 +1,26 @@
...
+  /**
+   * Determine if submission has scheduled email for specified handler.
+   *
+   * @param \Drupal\webform\WebformSubmissionInterface $webform_submission
+   *   A webform submission.
+   * @param \Drupal\node\Entity\Node $agencyComponent
+   *   The Agency Component node object.
+   *
+   * @return bool
+   *   TRUE if submission has scheduled email.
```
malikkotob: Copy/paste error (which I'm sure you know)

docroot/modules/custom/foia_webform/src/FoiaSubmissionServiceInterface.php
```diff
@@ -0,0 +1,26 @@
...
+   */
+  public function sendSubmissionToComponent(WebformSubmissionInterface $webform_submission, Node $agencyComponent);
```
malikkotob: `NodeInterface`

docroot/modules/custom/foia_webform/src/FoiaSubmissionApiService.php
```diff
@@ -0,0 +1,95 @@
+<?php
+
+namespace Drupal\foia_webform\Plugin\QueueWorker;
```
malikkotob: A lot of these have a `namespace` of `Drupal\foia_webform\Plugin\QueueWorker` but they should just be `Drupal\foia_webform`. I'd just double check the `namespace` throughout to confirm everything is correct.

docroot/modules/custom/foia_webform/foia_webform.services.yml
```diff
@@ -0,0 +1,20 @@
...
+    class: Drupal\foia_webform\FoiaSubmissionServiceFactory
+    arguments: ['@entity.manager', '@logger.channel.foia_webform']
+  foia_webform.foia_submission_queue_worker:
+    class: Drupal\foia_webform\Plugin\QueueWorker
+    arguments: ['@entity.manager', '@foia_webform.agency_lookup_service', '@foia_webform.foia_submission_service_api', '@foia_webform.foia_submission_service_email', '@foia_webform.foia_submission_service_factory']
+
+  logger.channel.foia_webform:
+    class: Drupal\Core\Logger\LoggerChannel
+    factory: logger.factory:get
+    arguments: ['foia_webform']
```
malikkotob: New line needed here

docroot/modules/custom/foia_webform/foia_webform.services.yml
```diff
@@ -0,0 +1,20 @@
...
+    class: Drupal\foia_webform\FoiaSubmissionServiceApi
+    arguments: ['@http_client', '@foia_webform.agency_lookup_service', '@logger.channel.foia_webform']
+  foia_webform.foia_submission_service_email:
+    class: Drupal\foia_webform\FoiaSubmissionServiceEmail
+    arguments: ['@foia_webform.agency_lookup_service', '@logger.channel.foia_webform']
+  foia_webform.foia_submission_service_factory:
+    class: Drupal\foia_webform\FoiaSubmissionServiceFactory
+    arguments: ['@entity.manager', '@logger.channel.foia_webform']
+  foia_webform.foia_submission_queue_worker:
+    class: Drupal\foia_webform\Plugin\QueueWorker
```
malikkotob: Missing the actual class here

docroot/modules/custom/foia_webform/src/AgencyLookupService.php
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * Class AgencyLookupService.
+ */
+class AgencyLookupService implements AgencyLookupServiceInterface {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getComponentFromWebform($webformId) {
+    $entityQueryService = \Drupal::service('entity.query');
```
malikkotob: Since we're using a service here, we should pass it in as an argument in the `services.yml` file like we do elsewhere.

docroot/modules/custom/foia_webform/src/AgencyLookupService.php
```diff
@@ -0,0 +1,34 @@
...
+
+    $node = ($nid) ? Node::load(reset($nid)) : NULL;
+    return $node;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getAgencyFromComponent(NodeInterface $agencyComponent) {
+    return $agencyComponent->get('field_agency')->getEntity();
```
malikkotob: Might want to incorporate some defensive programming checks in here in case the component isn't associated with an agency for some reason.

docroot/modules/custom/foia_webform/src/AgencyLookupServiceInterface.php
```diff
@@ -0,0 +1,34 @@
...
+  public function getComponentFromWebform($webformId);
+
+  /**
+   * Look up the Agency taxonomy term for a given Agency Component.
+   *
+   * @param \Drupal\node\NodeInterface $agencyComponent
+   *   The Agency Component node object.
+   *
+   * @return \Drupal\taxonomy\TermInterface
+   *   Returns the Agency taxonomy term object.
```
malikkotob: If we update this method to also return `NULL` in case the component isn't associated with an agency for whatever reason, this doc block will need updating as well to indicate that null is a valid return type, similar to what you did for the `getComponentFromWebform` method above.

docroot/modules/custom/foia_webform/src/FoiaSubmissionServiceApi.php
```diff
@@ -0,0 +1,155 @@
...
+use Drupal\file\Entity\File;
+use Drupal\node\NodeInterface;
+use Drupal\webform\WebformSubmissionInterface;
+use GuzzleHttp\ClientInterface;
+use Psr\Log\LoggerInterface;
+
+/**
+ * Class FoiaSubmissionServiceApi.
+ *
+ * @package Drupal\foia_webform\Plugin\QueueWorker
```
malikkotob: This should be removed

docroot/modules/custom/foia_webform/src/FoiaSubmissionServiceApi.php
```diff
@@ -0,0 +1,155 @@
...
+   *   The HTTP Guzzle client.
+   * @param \Drupal\foia_webform\AgencyLookupServiceInterface $agencyLookupService
+   *   The Agency Lookup service.
+   * @param \Psr\Log\LoggerInterface $logger
+   *   A logger instance.
+   */
+  public function __construct(ClientInterface $httpClient, AgencyLookupServiceInterface $agencyLookupService, LoggerInterface $logger) {
+    $this->httpClient = $httpClient;
+    $this->agencyLookUpService = $agencyLookupService;
+    $this->logger = $logger;
```
malikkotob: nice work.

docroot/modules/custom/foia_webform/src/FoiaSubmissionServiceApi.php
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function create(ConfigFactoryInterface $container, array $configuration, $plugin_id, $plugin_definition) {
```
malikkotob: I don't think you need this `create` method here.

docroot/modules/custom/foia_webform/src/FoiaSubmissionServiceApi.php
```diff
@@ -0,0 +1,155 @@
...
+      $container->get('foia_webform.agency_lookup_service'),
+      $container->get('logger.channel.foia_webform')
+    );
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function sendSubmissionToComponent(WebformSubmissionInterface $webformSubmission, NodeInterface $agencyComponent) {
+    $apiUrl = $agencyComponent->get('field_submission_api');
```
malikkotob: Maybe we should log an error if no url is set?

docroot/modules/custom/foia_webform/src/FoiaSubmissionServiceApi.php
```diff
@@ -0,0 +1,155 @@
...
+
+  /**
+   * Parse the Agency information from the Agency Component.
+   *
+   * @param \Drupal\node\NodeInterface $agencyComponent
+   *   The Agency Component node object.
+   */
+  public function getAgencyInfo(NodeInterface $agencyComponent) {
+    return [
+      'agency' => $this->agencyLookUpService->getAgencyFromComponent($agencyComponent),
```
malikkotob: This is going to return an entire entity, right? We probably just want the agency name here.

docroot/modules/custom/foia_webform/src/FoiaSubmissionServiceApi.php
```diff
@@ -0,0 +1,155 @@
...
+   *
+   * @return array
+   *   Returns the submission values as an array.
+   */
+  public function getSubmissionValues(WebformSubmissionInterface $webformSubmission) {
+    return $webformSubmission->getData();
+  }
+
+  /**
+   * Parse the Agency information from the Agency Component.
```
malikkotob: Maybe `Retrieve` or `Fetch` is more appropriate since we're not really parsing much here.

docroot/modules/custom/foia_webform/src/FoiaSubmissionServiceApi.php
```diff
@@ -0,0 +1,155 @@
...
+   *
+   * @param array $files
+   *   An array containing the file IDs for the file attachments.
+   */
+  public function getAttachmentData(array $files) {
+    if (!empty($files)) {
+      $fileData = [];
+      foreach ($files as $fid) {
+        $currentFile = File::load($fid);
+        print_r($currentFile);
```
malikkotob: what's this??? haha jk.

docroot/modules/custom/foia_webform/src/FoiaSubmissionServiceApi.php
```diff
@@ -0,0 +1,155 @@
...
+        $base64 = base64_encode(file_get_contents($currentFile->getFileUri()));
```
malikkotob: I tend to try and avoid alluding to the type or data specifics of the variable being stored. In this case we're holding file contents, and that would be the case even if we leveraged some sort of other encoding. So maybe this variable can be renamed `$fileContents`. There's a great book titled 'Clean Code' that talks about this. I see it as a form of 'Hungarian Notation'.

"...nowadays HN and other forms of type encoding are simply impediments. They make it harder to change the name or type of a variable, function, member or class. They make it harder to read the code. And they create the possibility that the encoding system will mislead the reader."

docroot/modules/custom/foia_webform/src/FoiaSubmissionServiceApi.php
```diff
@@ -0,0 +1,155 @@
...
+   * Return the FOIA form submission values as an array.
+   *
+   * @param \Drupal\webform\WebformSubmissionInterface $webformSubmission
+   *   The Webform submission values.
+   *
+   * @return array
+   *   Returns the submission values as an array.
+   */
+  public function getSubmissionValues(WebformSubmissionInterface $webformSubmission) {
+    return $webformSubmission->getData();
```
malikkotob: Are you planning on calling `getAttachmentData` so that we can manipulate this submission array as needed?

docroot/modules/custom/foia_webform/src/FoiaSubmissionServiceApi.php
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+  /**
+   * Get the metadata for the file attachments.
+   *
+   * @param array $files
+   *   An array containing the file IDs for the file attachments.
+   */
+  public function getAttachmentData(array $files) {
+    if (!empty($files)) {
```
malikkotob: Rather than passing an array of files here, we may want to consider passing the webform submission values array instead. Then we can iterate through that to find all fids and replace them with the file data you're retrieving below.

docroot/modules/custom/foia_webform/src/FoiaSubmissionServiceFactory.php
```diff
@@ -0,0 +1,36 @@
...
+  public function get(NodeInterface $agencyComponent) {
+    $submissionPreference = $agencyComponent->get('field_portal_submission_format');
+
+    switch ($submissionPreference) {
+      case 'api':
+        $serviceName = 'foia_webform.foia_submission_api_service';
+        break;
+
+      default:
+        \Drupal::logger('foia_webform')
```
malikkotob: Let's use the same logger pattern here.

docroot/modules/custom/foia_webform/src/Plugin/QueueWorker/FoiaSubmissionQueueWorker.php
```diff
@@ -19,29 +22,76 @@ class FoiaSubmissionQueueWorker extends QueueWorkerBase implements ContainerFact
...
+  public function processItem($data) {
+    $webformSubmission = $this->webformStorage->load($data->sid);
+
+    $agencyComponent = $this->agencyLookUpService->getComponentFromWebform($webformSubmission->getWebform()->id());
+    $submissionPreference = $agencyComponent->get('field_portal_submission_format');
+    if ($submissionPreference == 'api') {
+      $this->foiaSubmissionApiService->sendSubmissionToComponent($data, $agencyComponent);
+    }
+    else {
+      $this->foiaSubmissionEmailService->sendSubmissionToComponent($data, $agencyComponent);
```
malikkotob: Let's use the factory here instead of the if else.

docroot/modules/custom/foia_webform/src/FoiaSubmissionServiceApi.php
```diff
@@ -0,0 +1,158 @@
...
+   *
+   * @return array
+   *   Return the assemble request data in an array.
+   */
+  public function assembleRequestData(WebformSubmissionInterface $webformSubmission, NodeInterface $agencyComponent) {
+    // Get the webform submission values.
+    $submissionValues = $this->getSubmissionValues($webformSubmission);
+
+    // If there are files attached, load the files and add the file metadata.
+    if (isset($submissionValues['attachments_supporting_documentation'])) {
```
malikkotob: If we ever add an additional attachments field or replace this one, this code will stop working. I think what we probably have to do is search through the submission fields for all of the file fields. Maybe our best bet is to do that within `getAttachmentData` and to invoke `getAttachmentData` from within `getSubmissionValues` so that the code reads cleanly. Checkout our custom `WebformSubmissionResource` inside of `foia_api`, specifically the `getSubmittedFileAttachments` has code that shows how to dynamically find the file fields.

docroot/modules/custom/foia_webform/src/FoiaSubmissionServiceApi.php
```diff
@@ -0,0 +1,254 @@
...
+ * Class FoiaSubmissionServiceApi.
+ */
+class FoiaSubmissionServiceApi implements FoiaSubmissionServiceInterface {
+
+  /**
+   * The Agency Lookup service.
+   *
+   * @var \Drupal\foia_webform\AgencyLookupService
+   */
+  protected $agencyLookUpService;
```
malikkotob: Should probably spell `LookUp` with a lower case `u` for consistency with the actual class: `AgencyLookupService`

docroot/modules/custom/foia_webform/src/FoiaSubmissionServiceApi.php
```diff
@@ -0,0 +1,254 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function sendSubmissionToComponent(WebformSubmissionInterface $webformSubmission, WebformInterface $webform, NodeInterface $agencyComponent) {
+    $apiUrl = $agencyComponent->get('field_submission_api')->value;
+
+    if (!UrlHelper::isValid($apiUrl, TRUE)) {
+      $apiUrl = FALSE;
```
malikkotob: The `sendSubmissionToComponent` method should return true if the submission was sent, and false if it was not. As such, rather than setting this to false and then performing the `if ($apiUrl)` check below, we should just be returning false here.

docroot/modules/custom/foia_webform/src/FoiaSubmissionServiceApi.php
```diff
@@ -0,0 +1,254 @@
...
+      $fileKeys = $this->getFileAttachmentElementsOnWebform($webform);
+      if ($fileKeys) {
+        foreach ($fileKeys as $key) {
+          $files = $this->getAttachmentData($submissionValues[$key]);
+          $attachments = array_merge($files, $attachments);
+          unset($submissionValues[$key]);
+        }
+      }
+      if (!empty($attachments)) {
+        $submissionValues['attachments'] = $attachments;
```
malikkotob: The attachments should still be separated by field so that if we have two different file attachment fields that serve different purposes we retain that distinction.

docroot/modules/custom/foia_webform/src/FoiaSubmissionServiceApi.php
```diff
@@ -0,0 +1,254 @@
...
+    $fileContents = [];
+    if (!empty($files)) {
+      foreach ($files as $fid) {
+        $currentFile = File::load($fid);
+        $base64 = base64_encode(file_get_contents($currentFile->getFileUri()));
+        $fileContents[] = [
+          'content_type' => $currentFile->getMimeType(),
+          'filedata' => $base64,
+          'filename' => $currentFile->getFilename(),
+          'filesize' => $currentFile->getSize(),
```
malikkotob: the `getSize` method returns a string, mind if we cast it to an integer?
`(int)$currentFile->getSize();` should do it, but definitely test it out first.

docroot/modules/custom/foia_webform/src/FoiaSubmissionServiceFactoryInterface.php
```diff
@@ -0,0 +1,20 @@
...
+
+use Drupal\node\NodeInterface;
+
+/**
+ * Provides an interface for submitting the request using the preferred method.
+ */
+interface FoiaSubmissionServiceFactoryInterface {
+
+  /**
+   * Determines which method to use for submitting the request.
```
malikkotob: In addition to determining which method to use, it returns an instantiated service. Maybe we can change this to:
`Returns appropriate service for submitting the request based on component.`

docroot/modules/custom/foia_webform/src/AgencyLookupService.php
```diff
@@ -0,0 +1,51 @@
...
+    return ($nid) ? Node::load(reset($nid)) : NULL;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getAgencyFromComponent(NodeInterface $agencyComponent) {
+    $agencyTerm = NULL;
+    if (!$agencyComponent->get('field_agency')->isEmpty()) {
+      $agencyTerm = $agencyComponent->get('field_agency')->getEntity();
```
dangur: Instead of `getEntity` we'll want something like `Term::load($tid)`

malikkotob:
This has been rebased and merged into `develop`.

PR #85, opened 2017-10-06
DOJ-121: Tests that agency lookup service gets correct component from webform.

dangur:
As the last commit suggests, agency_compnent still needs field_request_submission_form.


malikkotob:
This has been rebased and merged into `develop`.

PR #87, opened 2017-10-11
DOJ-172: Add CORS config for POST requests.

adborden:
Adding settings in order to allow for POST requests from the front stage (for request submission).


malikkotob:
Thanks for this, @adborden. Would you mind also updating the `default.services.yml` file with the same contents, similar to what you did in the past?

adborden:
@malikkotob updated, thanks!

malikkotob:
All set!

PR #88, opened 2017-10-12
Updates to Agency Component Submission API fields.

ba66e77:
This PR:

1.  Adds a Submission API Shared Secret field to support the authentication method described in [18F's recommendations for portal integration](https://github.com/18F/beta.foia.gov/blob/master/agency-api-spec.md#authentication).

2.  Removes the existing Submission API text field and replaces it with a Submission API URL field which expects an external link to be provided as input.

**Note**
There is a [known issue in the Link module](https://www.drupal.org/node/2421001) which prevents the help text on link fields from displaying correctly.


config/default/core.entity_view_display.node.agency_component.teaser.yml
```diff
@@ -27,13 +27,17 @@ content:
...
+  field_agency_comp_telephone: true
   field_agency_component_id: true
   field_email: true
   field_foia_officers: true
+  field_misc: true
   field_portal_submission_format: true
   field_public_liaisons: true
   field_reading_rooms: true
   field_receiver: true
+  field_request_submission_form: true
```
malikkotob: @ba66e77 Why are we adding the request form to the teaser view here?

config/default/core.entity_form_display.node.agency_component.default.yml
```diff
@@ -17,8 +17,8 @@ dependencies:
     - field.field.node.agency_component.field_request_submission_form
     - field.field.node.agency_component.field_service_centers
     - field.field.node.agency_component.field_submission_address
-    - field.field.node.agency_component.field_submission_api
     - field.field.node.agency_component.field_submission_api_secret
+    - field.field.node.agency_component.field_submission_api_url
```
malikkotob: What're your thoughts on retaining the machine name `field_submission_api`? If we proceed with changing it to `field_submission_api_url`, we need to be sure we're modifying references to the field that explicitly look for `field_submission_api`. I know jsonapi config has references to that, I'm surprised it didn't get included in your exported config. Also, the api code coming down the pipeline looks for it as well.

config/default/field.field.node.agency_component.field_submission_api_url.yml
```diff
@@ -0,0 +1,23 @@
...
+langcode: en
+status: true
+dependencies:
+  config:
+    - field.storage.node.field_submission_api_url
+    - node.type.agency_component
+  module:
+    - link
+id: node.agency_component.field_submission_api_url
+field_name: field_submission_api_url
```
malikkotob: What're your thoughts on retaining the machine name field_submission_api? If we proceed with changing it to field_submission_api_url, we need to be sure we're modifying references to the field that explicitly look for field_submission_api. I know jsonapi config has references to that, I'm surprised it didn't get included in your exported config. Also, the api code coming down the pipeline looks for it as well.

config/default/core.entity_view_display.node.agency_component.teaser.yml
ba66e77: yeah, this was inadvertent.  Not sure how it got in there.  Will be fixed in next commit.

ba66e77:
@malikkotob, your previous comments should be resolved now.  Let me know what you think.

PR #90, opened 2017-10-13
DOJ-195: Tests that the api returns state/province values.

dangur:


docroot/modules/custom/webform_serialization/src/Normalizer/WebformNormalizer.php
```diff
@@ -23,6 +24,7 @@ public function getFields($entity, $bundle, ResourceType $resource_type) {
     if (!empty($enabled_public_fields['elements'])) {
       try {
         $parsed = Yaml::decode($enabled_public_fields['elements']);
+        $this->getSelectFields($parsed);
```
malikkotob: Typically, when we use `get` to prefix a method, we expect it to return something that we can act on. In this case we're passing in an argument by reference, and acting on it within the body of the method. Let's update the method to be called `populateSelectFieldsWithOptions`, rather than returning the select fields and acting on them here.

docroot/modules/custom/webform_serialization/src/Normalizer/WebformNormalizer.php
```diff
@@ -36,4 +38,20 @@ public function getFields($entity, $bundle, ResourceType $resource_type) {
     return $enabled_public_fields;
   }
 
+  /**
+   * Add the options for select items if the options are stored in config files.
```
malikkotob: "Replaces webform option machine names with fully rendered list of options."

docroot/modules/custom/webform_serialization/src/Normalizer/WebformNormalizer.php
```diff
@@ -36,4 +38,20 @@ public function getFields($entity, $bundle, ResourceType $resource_type) {
...
+   *
+   * @param array $formItems
```
malikkotob: Although these technically are form items, let's just call this variable `$elements`, since we're dealing with webforms.

docroot/modules/custom/webform_serialization/src/Normalizer/WebformNormalizer.php
```diff
@@ -36,4 +38,20 @@ public function getFields($entity, $bundle, ResourceType $resource_type) {
...
+   *   The already parsed form array.
```
malikkotob: Similar comment, let's update this lingo to be in line with "webform speak"

docroot/modules/custom/webform_serialization/src/Normalizer/WebformNormalizer.php
```diff
@@ -36,4 +38,20 @@ public function getFields($entity, $bundle, ResourceType $resource_type) {
...
+   */
+  private function getSelectFields(array &$formItems) {
```
malikkotob: Unless there's a strong reason to make this private, let's make it protected in case we ever extend this class and need to access this method from there.

docroot/modules/custom/webform_serialization/src/Normalizer/WebformNormalizer.php
```diff
@@ -36,4 +38,20 @@ public function getFields($entity, $bundle, ResourceType $resource_type) {
...
+    foreach ($formItems as $key => $element) {
+      if ($element['#type'] === 'select' && !is_array($element['#options'])) {
+        /** @var \Drupal\webform\Entity\WebformOptions $webformOptions */
+        $webformOptions = WebformOptions::getElementOptions($element);
+        $formItems[$key]['#options'] = $webformOptions;
```
malikkotob: Can we not just use the code below to set the options?
```
$element['#options'] = $webformOptions
```

docroot/modules/custom/webform_serialization/src/Normalizer/WebformNormalizer.php
```diff
@@ -36,4 +38,20 @@ public function getFields($entity, $bundle, ResourceType $resource_type) {
...
 
+  /**
+   * Add the options for select items if the options are stored in config files.
+   *
+   * @param array $formItems
+   *   The already parsed form array.
+   */
+  private function getSelectFields(array &$formItems) {
+    foreach ($formItems as $key => $element) {
+      if ($element['#type'] === 'select' && !is_array($element['#options'])) {
```
malikkotob: Nice

docroot/modules/custom/webform_serialization/tests/src/Functional/WebformNormalizationFormOptionsTest.php
```diff
@@ -0,0 +1,105 @@
...
+use Drupal\webform\Entity\Webform;
+use GuzzleHttp\Exception\ClientException;
+use GuzzleHttp\Exception\ServerException;
+use Drupal\Component\Serialization\Json;
+use Drupal\webform\Entity\WebformOptions;
+
+/**
+ * Tests for webform select options.
+ *
+ * @group jsonapi
```
malikkotob: Let's update this to `foiaapi` or something similar.

docroot/modules/custom/webform_serialization/tests/src/Functional/WebformNormalizationFormOptionsTest.php
```diff
@@ -0,0 +1,59 @@
...
+    'jsonapi',
+    'serialization',
+    'webform',
+    'webform_serialization',
+    'webform_template',
+    'user',
+  ];
+
+  /**
+   * Test the GET method.
```
malikkotob: "Tests that jsonapi returns fully rendered webform options."

docroot/modules/custom/webform_serialization/tests/src/Functional/WebformNormalizationFormOptionsTest.php
```diff
@@ -0,0 +1,59 @@
...
+   */
+  public function testRead() {
```
malikkotob: `testPopulateSelectFieldsWithOptions` to match the method we want to verify is working

docroot/modules/custom/webform_serialization/tests/src/Functional/WebformNormalizationFormOptionsTest.php
```diff
@@ -0,0 +1,59 @@
...
+use Drupal\webform\Entity\Webform;
+use Drupal\Component\Serialization\Json;
+use Drupal\webform\Entity\WebformOptions;
+
+/**
+ * Tests for webform select options.
+ *
+ * @group foiaapi
+ */
+class WebformNormalizationFormOptionsTest extends BrowserTestBase {
```
malikkotob: Typically, the test class name is the same as the class you're testing, with the word "Test" tacked onto the end. So in this case: `WebformNormalizerTest`.

docroot/modules/custom/webform_serialization/tests/src/Functional/WebformNormalizationFormOptionsTest.php
```diff
@@ -0,0 +1,59 @@
...
+  ];
+
+  /**
+   * Test the GET method.
+   */
+  public function testRead() {
+
+    // Create webform.
+    $this->webform = Webform::create(['id' => 'serialization_test']);
+    $this->webform->set('foia_template', [
```
malikkotob: should be able pass `1` instead of array, i.e.:
```
$this->webform->set('foia_template', 1);
```

PR #93, opened 2017-10-16
DOJ-192: Import annual processing data

dangur:


docroot/modules/custom/foia_migrate/src/data-groomer.php
```diff
@@ -300,3 +301,113 @@ function remove_departments_property_from_agencies(array &$agency_data) {
...
+ * @param array $latest_request_time_stats_array
+ *   An array of most recent year statistics.
+ */
+function remove_processing_from_expedited_fields(array $latest_request_time_stats_array) {
+  $keys = array_keys($latest_request_time_stats_array);
+  // Transform keys that begin w/'expedited_processing' to begin w/'expedited'.
+  if (in_array('expedited_processing_average_days', $keys) ||
+    in_array('expedited_processing_highest_days', $keys) ||
+    in_array('expedited_processing_lowest_days', $keys) ||
+    in_array('expedited_processing_median_days', $keys)
```
malikkotob: I don't think we need to do this additional processing. We should probably just update the `processing_data.yml` file to look for the fields with `processing` in the name, rather than changing the name to match what we have in the yml file. The less we do to the original data the better.
e.g:
```
    -
      name: expedited_average_days
      label: 'Expedited Average Days'
      selector: /latest_request_time_stats/expedited_average_days
```

docroot/modules/custom/foia_migrate/src/data-groomer.php
```diff
@@ -300,3 +301,113 @@ function remove_departments_property_from_agencies(array &$agency_data) {
...
+    'simple_median_days',
+  ];
+  foreach ($fields as $field) {
+    if (!array_key_exists($field, $latest_request_time_stats_array)) {
+      $latest_request_time_stats_array[$field] = '';
+    }
+  }
+  // Cleans up any array items with key of '0'.
+  if (array_key_exists('0', $latest_request_time_stats_array)) {
+    unset($latest_request_time_stats_array['0']);
```
malikkotob: Do you know why this is necessary?

docroot/modules/custom/foia_migrate/src/data-groomer.php
```diff
@@ -300,3 +301,113 @@ function remove_departments_property_from_agencies(array &$agency_data) {
...
+      $latest_data = extract_processing_data_from_agency($component->agency_name, $agency_data);
+      if ($latest_data) {
+        $component->latest_request_time_stats = $latest_data['data'];
+        $component->latest_request_time_stats_year = $latest_data['year'];
+      }
+    }
+
+    // Add empty values for any statistics missing data.
+    $latest_request_time_stats_array = (array) $component->latest_request_time_stats;
+    $component->latest_request_time_stats = (object) insert_empty_days_values($latest_request_time_stats_array);
```
malikkotob: Does the code not work as expected without the casts to `array` and `object`?

docroot/modules/custom/foia_migrate/src/data-groomer.php
```diff
@@ -300,3 +301,113 @@ function remove_departments_property_from_agencies(array &$agency_data) {
...
+    else {
+      $latest_data = extract_processing_data_from_agency($component->agency_name, $agency_data);
+      if ($latest_data) {
+        $component->latest_request_time_stats = $latest_data['data'];
+        $component->latest_request_time_stats_year = $latest_data['year'];
+      }
+    }
+
+    // Add empty values for any statistics missing data.
+    $latest_request_time_stats_array = (array) $component->latest_request_time_stats;
```
malikkotob: Preferable not to include the data type/structure in the variable name. Just `$latest_request_time_stats` would be sufficient here

tests/behat/features/AgencyComponent.feature
```diff
@@ -12,8 +12,42 @@ Feature: Agency Component Feature
...
+    And for 'Complex Lowest Days' I enter 'less than 1'
+    And for 'Complex Median Days' I enter '1'
+    And for 'Expedited Average Days' I enter '3.14'
+    And for 'Expedited Highest Days' I enter '0'
+    And for 'Expedited Lowest Days' I enter '3'
+    And for 'Expedited Median Days' I enter '9'
+    And for 'Simple Average Days' I enter '1'
+    And for 'Simple Highest Days' I enter '6'
+    And for 'Simple Lowest Days' I enter '1'
+    And for 'Simple Median Days' I enter '1'
```
malikkotob: From one of the technical requirements in the ticket:
> All processing data fields should be read-only for all users

While these tests do pass, non-admins should only be able to view the data, but should not be able to edit it. We need to update the permissions accordingly. In terms of the tests, they can verify that anonymous users (this implies that agency admins and managers can view it too, so no separate test really needed) can view the data and that non-admins can't edit the data.

docroot/modules/custom/foia_migrate/src/data-groomer.php
dangur: When I tried to send `$component->latest_request_time_stats` as an argument to the `insert_empty_days_values` function I got: `Catchable fatal error: Argument 1 passed to insert_empty_days_values() must be an instance of object, instance of stdClass given`

docroot/modules/custom/foia_migrate/src/data-groomer.php
dangur: I saw some array items with keys of 0 when I had debugging set up, so I removed those items. I'm not sure if it is necessary but seems like a good sanitary measure.

docroot/modules/custom/foia_migrate/src/data-groomer.php
malikkotob: Oh interesting. Hmm, if that's the case I would imagine this works?
```
$latest_request_time_stats = (array) $component->latest_request_time_stats;
$component->latest_request_time_stats = insert_empty_days_values($latest_request_time_stats);
```
(Removed the `(object)`)

docroot/modules/custom/foia_migrate/src/data-groomer.php
malikkotob: Oh weird. Yeah that sounds good.

config/default/user.role.agency_administrator.yml
```diff
@@ -34,17 +47,69 @@ permissions:
...
+  - 'edit own field_complex_median_days'
+  - 'edit own field_expedited_average_days'
+  - 'edit own field_expedited_highest_days'
+  - 'edit own field_expedited_lowest_days'
+  - 'edit own field_expedited_median_days'
+  - 'edit own field_request_data_year'
+  - 'edit own field_simple_average_days'
+  - 'edit own field_simple_highest_days'
+  - 'edit own field_simple_lowest_days'
+  - 'edit own field_simple_median_days'
```
malikkotob: Sorry @dangur, by 'non-admins' in [this comment](https://github.com/usdoj/foia-api/pull/93#discussion_r145154008), I meant everyone except the default Drupal administrator role. So even agency administrators should only be able to view these fields. These fields contain historical data, so we don't want to let anyone mess with it

malikkotob:
Covered by #95.

PR #100, opened 2017-10-31
DOJ-111: Add X-Api-Key header for CORS.

adborden:
api.data.gov has a bug where the pre-flight CORS request is denied (403
Forbidden), so the work around is to pass it through to the back end and have
the back end return the proper Access-Control headers.

https://github.com/NREL/api-umbrella/issues/391

### DIFF
--- a/docroot/sites/default/services.yml
+++ b/docroot/sites/default/services.yml
@@ -161,7 +161,9 @@ parameters:
   cors.config:
     enabled: true
     # Specify allowed headers, like 'x-allowed-header'.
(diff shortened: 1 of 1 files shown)


adborden:
This is needed to setup api.data.gov https://github.com/18F/beta.foia.gov/issues/111

adborden:
BTW, I left out the other yml file because it seems like the previous changes to it were removed, so I figured we don't want it to match docroot/sites/default/services.yml, let me know if I need to update it.

PR #104, opened 2017-11-07
DOJ-0: Moves FOIA Webform Kernel tests into base classes.

dangur:


docroot/modules/custom/foia_webform/tests/src/Kernel/FoiaSubmissionQueueHandlerTest.php
```diff
@@ -64,58 +27,15 @@ class FoiaSubmissionServiceQueueHandlerTest extends KernelTestBase {
...
-    'node',
-    'field',
-    'taxonomy',
-    'field_permissions',
-    'text',
-    'link',
-    'foia_request',
-    'options',
-  ];
+  public static $modules = ['node'];
```
malikkotob: `node` is covered in the `FoiaWebformKernelTestBase`

docroot/modules/custom/foia_webform/tests/src/Kernel/FoiaWebformApiKernelTestBase.php
```diff
@@ -0,0 +1,136 @@
...
+    'options',
+  ];
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function setUp() {
+    parent::setUp();
+    $this->installSchema('webform', ['webform']);
+    $this->installConfig(['webform', 'webform_template', 'foia_webform']);
```
malikkotob: I know we don't use `webform`, `webform_template`, and `foia_webform` in _every_ class, but I think we could move these to the `FoiaWebformKernelTestBase`, since it wouldn't be farfetched to assume that a class that extends the base that isn't testing API-related items would need some of these items.

docroot/modules/custom/foia_webform/tests/src/Kernel/AgencyLookupServiceTest.php
```diff
@@ -17,41 +15,26 @@
...
   protected function setUp() {
     parent::setUp();
-
     $this->installConfig(['system', 'taxonomy']);
     $this->installSchema('user', 'users_data');
-    $this->installEntitySchema('node');
-    $this->installEntitySchema('user');
-    $this->installEntitySchema('taxonomy_term');
-
   }
```
malikkotob: I think we do a lot of things manually in this test that we do in the `FoiaWebformApiKernelTestBase`:
1. Setup a mock webform (we setup a real webform in a base class that we could probably use)
2. Add the agency component content type (we do that in `setupAgencyComponent`, so maybe it can be moved to `FoiaWebformKernelTestBase`)
3. Set up `field_request_submission_form` (handled as part of `setupAgencyComponent`)
4. Create the agency vocabulary and add an agency taxonomy term (we do that in the `setup` method)
5. Create a test agency component that is associated to a webform and an agency (we do that as part of `setupAgencyComponent`)

dangur:
Accidentally closed and not sure I can reopen because I rebased. Moving this to [PR 107](https://github.com/usdoj/foia-api/pull/107)

PR #108, opened 2017-11-10
Doj 393 Miscellaneous Launch Prep Tasks

dangur:


docroot/robots.txt
```diff
@@ -14,48 +14,4 @@
...
-Disallow: /index.php/admin/
-Disallow: /index.php/comment/reply/
-Disallow: /index.php/filter/tips/
-Disallow: /index.php/node/add/
-Disallow: /index.php/search/
-Disallow: /index.php/user/password/
-Disallow: /index.php/user/register/
-Disallow: /index.php/user/login/
-Disallow: /index.php/user/logout/
+Disallow: /
```
malikkotob: Perfect! Except for the commit message :) "backstange" should be "backstage".

And actually, we don't even need to specify that here, because this codebase is only concerned with itself. So we could really just say "Prevent search engines from indexing pages." and that would be enough

config/default/views.view.search.yml
```diff
@@ -1,6 +1,6 @@
 uuid: a9ce491f-2b24-406d-8a94-025fe3bdd801
 langcode: en
-status: true
+status: false
```
malikkotob: Did you try disabling the core `search` module as a whole? Just curious, we probably don't have to do it.

docroot/.htaccess
```diff
@@ -69,6 +69,10 @@ AddEncoding gzip svgz
 <IfModule mod_rewrite.c>
   RewriteEngine on
 
+  # Forbid access to admin and node paths on the api2.foia.gov domain.
+  RewriteCond %{HTTP_HOST} api2.foia.gov
```
malikkotob: After thinking this through a bit more, instead of specifying which domains we want to block, maybe we should block all domains that aren't some variation of `admin.foia.gov`:

```
RewriteCond %{HTTP_HOST} !(.*)admin\.foia\.gov [NC]
```

The NC makes it case-insensitive.

config/default/views.view.search.yml
dangur: It looks like the search module had never been installed.

docroot/.htaccess
```diff
@@ -69,6 +69,10 @@ AddEncoding gzip svgz
...
+  RewriteRule ^(admin.*|node/.+) - [F]
```
malikkotob: Let's also block `/user`

config/default/views.view.search.yml
malikkotob: Oh interesting, you're right. This works then!

docroot/.htaccess
```diff
@@ -69,9 +69,10 @@ AddEncoding gzip svgz
 <IfModule mod_rewrite.c>
   RewriteEngine on
 
-  # Forbid access to admin and node paths on the api2.foia.gov domain.
+  # Forbid access to admin, node, and user paths on domains other than admin.foia.gov and local.dojfoia.gov.
+  RewriteCond %{HTTP_HOST} !local.dojfoia.gov
```
malikkotob: It would be nice if our local environment was setup to match the admin.foia.gov pattern, rather than having to make an exception here. Let me look into whether or not it's reasonable to change that at this stage.

docroot/.htaccess
```diff
@@ -69,9 +69,10 @@ AddEncoding gzip svgz
...
   RewriteCond %{HTTP_HOST} !(.*)admin\.foia\.gov$ [NC]
-  RewriteRule ^(admin.*|node/.+) - [F]
+  RewriteRule ^(admin.*|node/.+|user.*) - [F]
```
malikkotob: This will also block access to `/administration` and `/users`, which isn't the intention. Let's do something like this:
```
RewriteRule ^(admin($|/)|node($|/)|user($|/)) - [F]
```

Also, I just noticed we're not blocking `core/install.php` or `update.php`. We probably should, but that one wouldn't be restricted by domain. No one should really ever be accessing those. Let's do a rewrite rule before any of the rewrite condition statements:
```
RewriteRule ^(core/install\.php|update\.php) - [F]
```

tests/behat/features/Configuration.feature
```diff
@@ -0,0 +1,17 @@
...
+    Given I am logged in as a user with the "Administrator" role
+    When I am at "admin/modules"
+    Then the "Syslog" checkbox should be checked
+    And the "Database Logging" checkbox should not be checked
+
+  @api
+  Scenario: Only administrators should see error messages
+    Given I am logged in as a user with the "Administrator" role
+    When I am at "admin/config/development/logging"
+    Then I should see the radio button "None" selected
```
malikkotob: Nice additions!

docroot/sites/default/settings.php
```diff
@@ -766,3 +766,12 @@
...
 
+/**
+ * Disable caching on paths that begin with admin or node.
+ */
+if (isset($_GET['q'])) {
+  if (strpos($_GET['q'], 'admin') === 0 ||
+    strpos($_GET['q'], 'node') === 0) {
+    $config['page_cache_maximum_age'] = 0;
+  }
+}
```
malikkotob: What's this for?

dangur: To prevent caching of forbidden pages.

docroot/.htaccess
```diff
@@ -69,6 +69,20 @@ AddEncoding gzip svgz
...
+  # Forbid access to admin, node, and user paths on domains other than admin.foia.gov and local.dojfoia.gov.
+  RewriteCond %{ENV:AH_SITE_ENVIRONMENT} ^(prod|test|dev)$
+  RewriteCond %{HTTP_HOST} !(.*)admin\.foia\.gov$ [NC]
+  RewriteRule ^((admin|node|user)($|/)) - [F]
+
+  # Redirect all HTTP server traffic to HTTPS.
+  RewriteCond %{ENV:AH_SITE_ENVIRONMENT} ^(prod|test|dev)$
+  RewriteCond %{HTTPS} off
+  RewriteCond %{HTTP:X-Forwarded-Proto} !https
+  RewriteRule ^(.*)$ https://%{HTTP_HOST}%{REQUEST_URI} [NE,L,R=301]
```
malikkotob: What does the `NE` flag do?

docroot/.htaccess
```diff
@@ -69,6 +69,20 @@ AddEncoding gzip svgz
 <IfModule mod_rewrite.c>
   RewriteEngine on
 
+  # Forbid access to install and update files.
+  RewriteRule ^(core/install|update)\.php - [F]
+
+  # Forbid access to admin, node, and user paths on domains other than admin.foia.gov and local.dojfoia.gov.
+  RewriteCond %{ENV:AH_SITE_ENVIRONMENT} ^(prod|test|dev)$
```
malikkotob: Is it possible to check if the variable is empty, rather than hardcoding environments? Just in case more environments are added down the road?

docroot/.htaccess
```diff
@@ -69,6 +69,20 @@ AddEncoding gzip svgz
 <IfModule mod_rewrite.c>
   RewriteEngine on
 
+  # Forbid access to install and update files.
+  RewriteRule ^(core/install|update)\.php - [F]
+
+  # Forbid access to admin, node, and user paths on domains other than admin.foia.gov and local.dojfoia.gov.
```
malikkotob: Comment needs updating since we're not specifying our local domain any longer. Maybe we can keep it generic in terms of the paths as well, in case we add more down the road.

docroot/.htaccess
```diff
@@ -69,6 +69,20 @@ AddEncoding gzip svgz
...
+  RewriteCond %{ENV:AH_SITE_ENVIRONMENT} ^(prod|test|dev)$
+  RewriteCond %{HTTP_HOST} !(.*)admin\.foia\.gov$ [NC]
+  RewriteRule ^((admin|node|user)($|/)) - [F]
+
+  # Redirect all HTTP server traffic to HTTPS.
+  RewriteCond %{ENV:AH_SITE_ENVIRONMENT} ^(prod|test|dev)$
```
malikkotob: Same comment as above, can we check if the variable is empty?

docroot/.htaccess
dangur: No Escape - doesn't convert special characters to hexcode (e.g. & => %26) potentially resulting in double encoding or 404 not found error conditions.

docroot/.htaccess
```diff
@@ -69,6 +69,20 @@ AddEncoding gzip svgz
 <IfModule mod_rewrite.c>
   RewriteEngine on
 
+  # Forbid access to install and update files.
+  RewriteRule ^(core/install|update)\.php - [F]
```
malikkotob: This shouldn't happen locally. We can just check to make sure we're on cloud with the same method you used below:
```
RewriteCond %{ENV:AH_SITE_ENVIRONMENT} ^.+$
```

docroot/sites/default/settings.php
```diff
@@ -749,6 +749,32 @@
   'bower_components',
 ];
 
+/**
+ * Memcache configuration.
+ */
+if ($is_ah_env) {
+  switch ($ah_env) {
+    case 'test':
```
malikkotob: Let's add `dev` here as well

dangur: Good catch, that should be done now

PR #113, opened 2017-11-21
DOJ-446 Add Confirmation ID to Email and API

dangur:


docroot/modules/custom/foia_webform/src/FoiaSubmissionServiceApi.php
```diff
@@ -122,7 +122,8 @@ protected function assembleRequestData(FoiaRequestInterface $foiaRequest) {
 
     $apiVersion = ['version' => FoiaSubmissionServiceInterface::VERSION];
     $foiaRequestId = ['request_id' => $foiaRequest->id()];
-    $submissionValues = array_merge($foiaRequestId, $apiVersion, $agencyInfo, $formValues);
+    $confirmationId = ['confirmation_id' => $foiaRequest->get('field_webform_submission_id')->value];
```
malikkotob: On line 118 we have:
```
$formValues = $this->getSubmissionValues($foiaRequest);
```

Within that method (`getSubmissionValues`), we grab the value for the submission ID so that we can load the webform submission. I wonder if we could just modify that method to include the `confirmation_id` as part of the array that `getSubmissionValues` returns so that we don't have to fetch the webform submission ID twice.

docroot/modules/custom/foia_webform/src/Plugin/WebformHandler/FoiaEmailWebformHandler.php
```diff
@@ -74,6 +74,7 @@ public function getEmailMessage($foiaRequestId, WebformSubmissionInterface $webf
     // Get form submission contents.
     $submissionContents = $webformSubmission->getData();
     $this->listFileAttachmentNamesInSubmission($submissionContents);
+    $submissionContents['confirmation_id'] = $webformSubmission->id();
```
malikkotob: This is better than what I did to add the request id. I did that within the `formatSubmissionContentsAsTable` method. Do you mind refactoring that to match the approach you took here?

PR #116, opened 2017-12-04
DOJ 453 Update request creation logic

dangur:


docroot/modules/custom/foia_request/src/Entity/FoiaRequestInterface.php
```diff
@@ -28,6 +28,11 @@
    */
   const STATUS_FAILED = 2;
 
+  /**
+   * Request pending file attachment virus scan.
+   */
+  const STATUS_SCAN = 3;
+
```
malikkotob: I think we need to update the `getRequestStatus` and `setRequestStatus` doc blocks to include this new status.

docroot/modules/custom/foia_webform/src/Plugin/WebformHandler/FoiaSubmissionQueueHandler.php
```diff
@@ -78,6 +78,11 @@ protected function createFoiaRequest(WebformSubmissionInterface $webformSubmissi
       $foiaRequest->set('field_requester_email', $requesterEmailAddress);
     }
 
+    $fileAttachment = $webformSubmission->getElementData('attachments_supporting_documentation');
```
malikkotob: Unfortunately we can't hardcode the `attachments_supporting_documentation` element here, since it's possible that a webform will have an additional attachment field.

dangur: @malikkotob My understanding is that the following conditional will set the entire request to be queued for scanning. Then during the request scanning process, individual files will have their statuses set appropriately. Thus, we would not need to differentiate between requests that have one attachment and those that have more than one attachment here.

malikkotob: @dangur Hmm, yeah, that's true, but I think the issue is that this field could technically be deleted. Also, even if it isn't deleted, if a different attachment field also existed on the form, and the user only uploaded an attachment to the other field, those files would not be scanned.

PR #118, opened 2017-12-06
WIP: DOJ 452 Extend file entity

dangur:


dangur:
[PR 120](https://github.com/usdoj/foia-api/pull/120) Will be used instead.

PR #121, opened 2017-12-15
DOJ-457: Update submission queue handler

Tom-Camp:
Moving `add to queue` to its own class and only adding "clean" submissions to queue.

### DIFF
--- a/docroot/modules/custom/foia_webform/src/Plugin/WebformHandler/FoiaSubmissionQueueHandler.php
+++ b/docroot/modules/custom/foia_webform/src/Plugin/WebformHandler/FoiaSubmissionQueueHandler.php
@@ -78,6 +78,11 @@ class FoiaSubmissionQueueHandler extends EmailWebformHandler {
       $foiaRequest->set('field_requester_email', $requesterEmailAddress);
     }
 
+    $fileAttachment = $webformSubmission->getElementData('attachments_supporting_documentation');
+    if ($fileAttachment) {
+      $foiaRequest->setRequestStatus(FoiaRequestInterface::STATUS_SCAN);
+    }
+
     $foiaRequest->save();
     return $foiaRequest;
   }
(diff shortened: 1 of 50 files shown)


dangur:
I think we should merge these PRs into the `file-scanning` branch instead of the `develop` branch.

docroot/modules/custom/foia_webform/src/Plugin/WebformHandler/FoiaSubmissionQueueHandler.php
```diff
@@ -31,7 +33,11 @@ public function postSave(WebformSubmissionInterface $webformSubmission, $update
       $componentAssociatedToWebform = $this->getComponentAssociatedToWebform($webformSubmission);
       if ($componentAssociatedToWebform) {
         $foiaRequest = $this->createFoiaRequest($webformSubmission, $componentAssociatedToWebform);
-        $this->queueFoiaRequest($foiaRequest);
+      }
+
+      if ($foiaRequest && $foiaRequest->getRequestStatus() === 0) {
```
dangur: Instead of `0`, can we use the constant `FoiaRequestInterface::STATUS_QUEUED` for clarity?

docroot/modules/custom/foia_webform/src/Plugin/WebformHandler/FoiaSubmissionQueueHandler.php
```diff
@@ -78,35 +84,71 @@ protected function createFoiaRequest(WebformSubmissionInterface $webformSubmissi
...
-
-    // @var QueueInterface $queue
-    $foiaSubmissionsQueue = $queueFactory->get('foia_submissions');
-    $submission = new \stdClass();
-    $submission->id = $foiaRequest->id();
+  protected function fileAttachmentSubmitted(WebformSubmissionInterface $webformSubmission) {
+    $webform = $webformSubmission->getWebform();
+    if ($webform->hasManagedFile()) {
+      $fileAttachmentElementsOnWebform = $this->getFileAttachmentElementsOnWebform($webform);
+      return $this->fileAttachmentsExist($fileAttachmentElementsOnWebform, $webformSubmission);
```
dangur: Should this return TRUE like on line 148?

Tom-Camp: `$this->fileAttachmentsExists()` returns TRUE or FALSE.

docroot/modules/custom/foia_webform/src/Plugin/WebformHandler/FoiaSubmissionQueueHandler.php
Tom-Camp: Yep. Sounds like a good idea.

PR #122, opened 2017-12-18
DOJ-455 Create file entity updater

dangur:


composer.json
```diff
@@ -26,7 +26,8 @@
         "drupal/swiftmailer": "^1.0@beta",
         "drupal/rules": "^3.0@alpha",
         "drupal/autologout": "^1.0",
-        "drupal/file_entity": "^2.0@beta"
+        "drupal/file_entity": "^2.0@beta",
+	"drupal/clamav": "^1.0"
```
malikkotob: Hmm, what is this commit for? We're not actually using the `clamav` module itself, right?

docroot/modules/custom/foia_file/foia_file.drush.inc
```diff
@@ -0,0 +1,43 @@
...
+ * Implements hook_drush_command().
+ */
+function foia_file_drush_command() {
+  $items = [];
+
+  $items['file-entity-updater'] = [
+    'description' => 'Updates file entities virus scan statuses.',
+    'core' => ['8+'],
+    'arguments' => [
+      'fully_qualified_fine_name' => 'A filename that includes the complete absolute path on the server.',
```
malikkotob: Typo: should be `file_name` instead of `fine_name`

docroot/modules/custom/foia_file/foia_file.drush.inc
```diff
@@ -0,0 +1,43 @@
...
+ * FOIA Api module drush commands.
+ */
+
+/**
+ * Implements hook_drush_command().
+ */
+function foia_file_drush_command() {
+  $items = [];
+
+  $items['file-entity-updater'] = [
```
malikkotob: I think drush commands are typically named in the form of actions, so this would be `file-entity-update`, rather than `updater`

composer.json
dangur: Correct, it was deleted in a subsequent commit.

docroot/modules/custom/foia_file/foia_file.drush.inc
dangur: 👍 I'll replace all instances of `updater` with `update`

docroot/modules/custom/foia_file/foia_file.drush.inc
dangur: Fine catch, 😄  I ended up pulling that out in a subsequent commit and just passing in the entire output of the scan results lines.

dangur:
These comments should all be resolved. I pushed up a new commit and am deploying to UAT now.

docroot/modules/custom/foia_file/foia_file.drush.inc
```diff
@@ -0,0 +1,108 @@
...
+    $status = $wordsArray[0];
+
+    // Get relative file path.
+    $scanLength = strlen($scan);
+    $statusLength = strlen($status);
+    $fileLength = $scanLength - $statusLength - 2;
+    $fileNameWithVirusName = substr($scan, 0, $fileLength);
+
+    // Trim virus description.
+    preg_match('/(\.(pdf|txt|doc|docx|fodt|odt|gif|jpg|png))(: .*)?$/', $fileNameWithVirusName, $fileExtension);
```
malikkotob: Might want to test this with files that have weird names. I'm also thinking we'll need to store these file extensions in config in case they ever change. Or, we might be able to grab the list of allowed extensions from existing config we have committed from our webform configuration.

docroot/modules/custom/foia_file/foia_file.drush.inc
```diff
@@ -0,0 +1,108 @@
...
+      }
+      elseif ($status === "Removed.") {
+        $fileEntity->set('field_virus_scan_status', 'virus');
+        $message .= "File {$webformDir}{$fileNameWithExtension} Entity ID {$fid} set to virus.\n";
+      }
+      $fileEntity->save();
+    }
+    else {
+      if ($fileEntity) {
+        $message .= "File {$webformDir}{$fileNameWithExtension} Entity ID {$fid} has already been scanned {$fileNameWithExtension}.\n";
```
malikkotob: It may have been scanned, but it also may have no status set, for whatever reason. We should talk about how to handle that scenario. Might be risky to assume that it has been scanned unless it has a status explicitly indicating that it has been.

docroot/modules/custom/foia_file/foia_file.drush.inc
```diff
@@ -0,0 +1,108 @@
...
+      }
+    }
+
+  }
+
+  $endTime = microtime(TRUE);
+
+  $executionTime = ($endTime - $startTime);
+
+  $message .= "File Entity update completed in {$executionTime} seconds.";
```
malikkotob: Nice, I like that you thought to include this.

docroot/modules/custom/foia_file/foia_file.drush.inc
```diff
@@ -75,7 +75,7 @@ function drush_foia_file_file_entity_updater($scanResult) {
 
     // Update file entity based upon status.
     $fileEntity = FileEntity::load($fid);
-    if ($fileEntity && $fileEntity->get('field_virus_scan_status') === 'scan') {
+    if ($fileEntity && $fileEntity->get('field_virus_scan_status')->getString() === 'scan') {
```
malikkotob: I think `->value` should work as well, which is what we use elsewhere. Might want to try that so that we're consistent in our approach to accessing field values. If that doesn't work let's use the `getString` method.

docroot/modules/custom/foia_file/drush/foia_file.drush.inc
```diff
@@ -0,0 +1,198 @@
...
+
+  // Temporarily delete similar paths of the webform dir during processing.
+  $trimmedFiles = str_replace($webformDir, '', $scanOutput);
+
+  // Split string into lines. @todo change to preg split \R?
+  $scans = explode("\n", $trimmedFiles);
+
+  // Get allowed extensions.
+  $config = \Drupal::config('file_entity.settings');
+  $extensions = $config->get('default_allowed_extensions');
```
malikkotob: Let's talk this through. What happens if this code is passed a file that is not one of the above extensions? For example, as of right now webform submissions can contain `odf` files, but that isn't one of the extensions that will get returned here. Do we know how the code will handle that scenario? I wonder if this code needs to be more generic to be able to account for any file extension

malikkotob:
The `foia_file.settings.yml` file is empty, can it be deleted?

docroot/modules/custom/foia_file/drush/foia_file.drush.inc
```diff
@@ -0,0 +1,198 @@
...
+      $msg .= 'This is not expected behavior.';
+      \Drupal::logger('foia_file')->warning($msg);
+    }
+
+    // Split the scan line into filename and status.
+    preg_match('/[^ ]*$/', $scan, $wordsArray);
+    // Set status to the last word of the line.
+    $status = $wordsArray[0];
+
+    $validStates = ['OK', 'Removed.', 'FOUND'];
```
malikkotob: Maybe we should rename this to `$validStatuses` to align with our usage of the `$status` variable.

docroot/modules/custom/foia_file/drush/foia_file.drush.inc
```diff
@@ -0,0 +1,198 @@
...
+    if (!in_array($status, $validStates)) {
+      $msg = 'An invalid status was detected in the virus scan output.';
+      $msg .= 'Virus scanning will not continue until this is resolved!!!';
+      \Drupal::logger('foia_file')->error($msg);
+      return drush_set_error('POLICY_PM_DENY', dt($msg));
```
malikkotob: I'm thinking we should talk through this scenario. So if a file at the beginning of the batch kept on returning a status that our system does not recognize, it would halt virus scanning for the entire system. As the system stands, there are no alerts in place, so processing for requests with attachments would essentially come to a screeching halt until someone noticed. What if we skipped invalid statuses instead of halting the process outright?

malikkotob: I just encountered an `Empty file` status as well.

So I worry there are non-fatal statuses that we aren't expecting that we can't safely assume should halt all the things.

docroot/modules/custom/foia_file/drush/foia_file.drush.inc
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    // Get relative file path.
+    $scanLength = strlen($scan);
+    $statusLength = strlen($status);
+    $fileLength = $scanLength - $statusLength - 2;
```
malikkotob: Add comment to explain why we need the `-2`

docroot/modules/custom/foia_file/drush/foia_file.drush.inc
```diff
@@ -0,0 +1,198 @@
...
+    $fileNameWithVirusName = substr($scan, 0, $fileLength);
+
+    // Trim virus description.
+    preg_match('/(\.(' . $pipeDelimitedExtensions . '))(: .*)?$/', $fileNameWithVirusName, $fileExtension);
```
malikkotob: No longer need to worry about virus descriptions since we'll skip lines with a status of "Found."

docroot/modules/custom/foia_file/drush/foia_file.drush.inc
```diff
@@ -0,0 +1,198 @@
...
+      $msg .= 'Virus scanning will not continue until this is resolved!!!';
+      \Drupal::logger('foia_file')->error($msg);
+      return drush_set_error('POLICY_PM_DENY', dt($msg));
+    }
+
+    // Get relative file path.
+    $scanLength = strlen($scan);
+    $statusLength = strlen($status);
+    $fileLength = $scanLength - $statusLength - 2;
+    $fileNameWithVirusName = substr($scan, 0, $fileLength);
```
malikkotob: This will be changed to `$fileName`

docroot/modules/custom/foia_file/drush/foia_file.drush.inc
```diff
@@ -0,0 +1,175 @@
...
+  foreach ($scans as $scan) {
+
+    // Split the scan line into filename and status.
+    preg_match('/[^ ]*$/', $scan, $wordsArray);
+    // Set status to the last word of the line.
+    $status = $wordsArray[0];
+
+    $validStatuses = ['OK', 'Removed.', 'Empty file'];
+    if (!in_array($status, $validStatuses)) {
+      \Drupal::logger('foia_file')->critical("An invalid status was detected in the virus scan output.");
```
malikkotob: Should we be considering this a critical error, or even invalid at all? I'm thinking there could be another status that we don't know about that is valid, but unexpected. The only thing we could say about it for sure is that it's `unexpected`, but without actually seeing what the scanner spit out, it's difficult to classify it as invalid or valid.

docroot/modules/custom/foia_file/drush/foia_file.drush.inc
```diff
@@ -0,0 +1,175 @@
...
+      $status = 'invalid';
+    }
+
+    // Get relative file path.
+    $scanLength = strlen($scan);
+    $statusLength = strlen($status);
+    // Length of the file name is the remainder of the scan minus the colon and
+    // space characters.
+    $fileLength = $scanLength - $statusLength - 2;
```
malikkotob: As mentioned in the comment above, this is the length of the file name, as opposed to what the variable is named, which is `fileLength`. Maybe we should rename this to more clearly indicate that we're talking about the length of the file name.

docroot/modules/custom/foia_file/drush/foia_file.drush.inc
```diff
@@ -0,0 +1,175 @@
...
+        "The file @fileName was detected on the filesystem and a virus was detected in it. The file should have been deleted by the scanner.",
+        ['@fileName' => $fileName]
+      );
+    }
+
+    // Determine file entity id based upon file path & name.
+    $query = \Drupal::entityQuery('file')
+      ->condition('uri', "private://webform/{$fileName}");
+    $fids = $query->execute();
+    $fid = array_values($fids)[0];
```
malikkotob: I think this might cause an error if `$fids` is empty.

docroot/modules/custom/foia_file/drush/foia_file.drush.inc
```diff
@@ -0,0 +1,174 @@
...
+
+    // @todo Validate that only one item in fids array.
+
+    // @todo Nice to have >> verify that submission id is same as file path.
+
+    // Update file entity based upon status.
+    if ($fid) {
+      $fileEntity = FileEntity::load($fid);
```
malikkotob: I'm not sure if this is guaranteed to be a file of type `FileEntity`.  We may need to use `File::load` just in case.

docroot/modules/custom/foia_file/drush/foia_file.drush.inc
```diff
@@ -0,0 +1,174 @@
...
+    }
+    else {
+      $fileEntity = NULL;
+    }
+    $doNotScanStatuses = ['clean', 'virus', 'invalid'];
+    if ($fileEntity && !in_array($fileEntity->get('field_virus_scan_status')->getString(), $doNotScanStatuses)) {
```
malikkotob: Since we can't be positive that the file will be of type `FileEntity`, we also can't assume it has the `field_virus_scan_status` on it. So before we do any updates on the file, we should check for the existence of the field. If it doesn't exist we can skip over it.

malikkotob: Also, I think I mentioned this previously, but I know there's been a lot of back and forth. I think `->value` should work as well instead of `getString()`, which is what we use elsewhere. Might want to try that so that we're consistent in our approach to accessing field values. If that doesn't work let's use the `getString` method.

docroot/modules/custom/foia_file/drush/foia_file.drush.inc
```diff
@@ -0,0 +1,174 @@
...
+        $fileEntity->set('field_virus_scan_status', 'clean');
+        \Drupal::logger('foia_file')->info(
+          "File @webformDir@fileName Entity ID @fid set to clean.",
+          [
+            '@webformDir' => $webformDir,
+            '@fileName' => $fileName,
+            '@fid' => $fid,
+          ]
+        );
+      }
```
malikkotob: Since most files are probably going to be clean (I hope, at least..) we probably don't have to log this. We may just end up filling up the logs.

docroot/modules/custom/foia_file/drush/foia_file.drush.inc
```diff
@@ -0,0 +1,174 @@
...
+      elseif ($status === "Removed.") {
+        $fileEntity->set('field_virus_scan_status', 'virus');
+        \Drupal::logger('foia_file')->warning(
+          "A virus was detected in the file @webformDir@fileName associated with the Entity ID @fid. The file has been deleted.",
+          [
+            '@webformDir' => $webformDir,
```
malikkotob: Is there a trailing slash in the `$webformDir` variable?

docroot/modules/custom/foia_file/drush/foia_file.drush.inc
```diff
@@ -0,0 +1,174 @@
...
+            '@fileName' => $fileName,
+            '@fid' => $fid,
+          ]
+        );
+        \Drupal::logger('foia_file')->notice(
```
malikkotob: I don't think we need both of these log messages.

docroot/modules/custom/foia_file/drush/foia_file.drush.inc
```diff
@@ -0,0 +1,174 @@
...
+    // Update file entity based upon status.
+    if ($fid) {
+      $fileEntity = FileEntity::load($fid);
+    }
+    else {
+      $fileEntity = NULL;
+    }
+    $doNotScanStatuses = ['clean', 'virus', 'invalid'];
+    if ($fileEntity && !in_array($fileEntity->get('field_virus_scan_status')->getString(), $doNotScanStatuses)) {
+      if ($status === "OK") {
```
malikkotob: We should also be marking empty files as clean

docroot/modules/custom/foia_file/drush/foia_file.drush.inc
```diff
@@ -0,0 +1,174 @@
...
+    // @todo Nice to have >> verify that submission id is same as file path.
+
+    // Update file entity based upon status.
+    if ($fid) {
+      $fileEntity = FileEntity::load($fid);
+    }
+    else {
+      $fileEntity = NULL;
+    }
+    $doNotScanStatuses = ['clean', 'virus', 'invalid'];
```
malikkotob: I don't think `invalid` will ever be the status

docroot/modules/custom/foia_file/drush/foia_file.drush.inc
```diff
@@ -0,0 +1,174 @@
...
+            '@fid' => $fid,
+          ]
+        );
+      }
+      $fileEntity->save();
+    }
+    else {
+      if ($fileEntity) {
+        \Drupal::logger('foia_file')->info(
+          "File @webformDir@fileName Entity ID @fid has already been scanned.",
```
malikkotob: Let's not log this. It might fill up the logs if we don't get the archiving work done.

docroot/modules/custom/foia_file/drush/foia_file.drush.inc
```diff
@@ -0,0 +1,174 @@
...
+          [
+            '@webformDir' => $webformDir,
+            '@fileName' => $fileName,
+            '@fid' => $fid,
+          ]
+        );
+      }
+      else {
+        \Drupal::logger('foia_file')->warning(
+          "There is no entity associated with @fileName.",
```
malikkotob: Same deal here, maybe we don't have to log this. It's nothing too spectacular. It's also not something we should really ever encounter, since files within the webform directory should always (theoretically) be associated with a file entity.

docroot/modules/custom/foia_file/drush/foia_file.drush.inc
```diff
@@ -0,0 +1,174 @@
...
+    $fileNameLength = $scanLength - $statusLength - 2;
+    $fileName = substr($scan, 0, $fileNameLength);
+
+    // Validate path, filename, and extension.
+    $pathParts = pathinfo($fileName);
+    foreach ($pathParts as $pathPart) {
+      if (empty($pathPart)) {
+        $status = 'invalid';
+      }
+    }
```
malikkotob: I'm not sure this check is necessary, since we don't really act on `invalid` statuses. We're probably also not 100% familiar with how this function behaves, so I'm hesitant to add in a check using it unless we know exactly what we're looking for. A file might not have an extension, for example, so that path part might be empty. I'm not sure we'd want to consider that invalid. We could just try to look up the corresponding file entity regardless.

docroot/modules/custom/foia_file/drush/foia_file.drush.inc
```diff
@@ -0,0 +1,174 @@
...
+
+    // Log any discrepancies between filesystem and entities.
+    if ($status === 'OK' && !is_file($fileName)) {
+      \Drupal::logger('foia_file')->warning(
+        "The file @fileName was not detected on the filesystem and was marked as clean by the virus scanner.",
+        ['@fileName' => $fileName]
+        );
```
malikkotob: spacing off for the closing `);`

malikkotob:
@dangur I think we can delete the `foia_file.settings.yml` file

docroot/modules/custom/foia_file/drush/foia_file.drush.inc
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    $relativeFileName = preg_replace('/: (OK|Empty file|Removed\.)$/', '', $scan);
+    $absoluteFileName = $webformDir . $relativeFileName;
+    $pathParts = pathinfo($absoluteFileName);
+
+    // Log any discrepancies between filesystem and entities.
+    if (in_array($scanStatus, $cleanScanStatuses) && !file_exists($absoluteFileName)) {
+      \Drupal::logger('foia_file')->warning(
+        "The file @absoluteFileName was not detected on the filesystem but was marked as clean by the virus scanner.",
```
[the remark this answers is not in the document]
dangur: I just saw this comment. This seems to happen in the test environments when file entities do not necessarily match the filesystem. In real life, this might happen if a file is deleted or corrupted on the filesystem. Ideally, a notification similar to that of when a file is intentionally deleted should be sent to the agency component so they can know the file is missing.

docroot/modules/custom/foia_file/drush/foia_file.drush.inc
```diff
@@ -0,0 +1,158 @@
...
+    // @todo Validate that only one item in fids array.
+
+    // @todo Nice to have >> verify that submission id is same as file path.
+
+    // Update file entity based upon scanStatus.
+    if ($fid) {
+      $file = File::load($fid);
+    }
+    else {
+      $file = NULL;
```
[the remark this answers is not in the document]
dangur: I just saw this comment too. It might be an error condition, but it is most likely when the scan line has a status of `FOUND`.

tests/behat/features/AgencyComponent.feature
```diff
@@ -12,6 +12,8 @@ Feature: Agency Component Feature
       | title                   | field_agency_comp_abbreviation | field_agency_comp_telephone | Request Submission Form | field_request_data_year | field_complex_average_days | field_complex_highest_days | field_complex_lowest_days | field_complex_median_days | field_expedited_average_days | field_expedited_highest_days | field_expedited_lowest_days | field_expedited_median_days | field_simple_average_days | field_simple_highest_days | field_simple_lowest_days | field_simple_median_days |
       | A Test Agency Component | TST                            | (555) 555-5555              | a_test_webform          | 2016                    | 1                          | 1                          | Less than 1               | 1                         | 3.14                         | 0                            | 3                           | 9                           | 1                         | 6                         | 1                        | 1                        |
     And I am at 'admin/content'
+    And for 'Title' I enter 'A Test Agency Component'
```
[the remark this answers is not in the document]
dangur: This is supposed to be part of the PR. Because there are so many agency components now, we are not guaranteed to see our test agency component on the first page.

PR #123, opened 2017-12-19
DOJ-456: Queueing clean requests.

Tom-Camp:


docroot/modules/custom/foia_request/foia_request.drush.inc
```diff
@@ -0,0 +1,158 @@
+<?php
+
+/**
+ * @file
+ * Drush command to update FOIA request submissions.
+ */
+
+use Drupal\webform\entity\WebformSubmission;
+use Drupal\file\Entity\File;
```
dangur: We should use file_entity instead of file.

docroot/modules/custom/foia_request/foia_request.drush.inc
```diff
@@ -0,0 +1,158 @@
...
+/**
+ * Check the status of files to see if they've been scanned.
+ *
+ * @param array $file_ids
+ *   An array containing file IDs.
+ */
+function foia_request_check_file_statuses(array $file_ids) {
+  $ready_to_queue = FALSE;
+  foreach ($file_ids as $file_id) {
+    $file = File::load($file_id);
```
dangur: We should use FileEntity instead of File

docroot/modules/custom/foia_request/foia_request.drush.inc
```diff
@@ -0,0 +1,158 @@
...
+        return FALSE;
+      }
+    }
+  }
+  return $ready_to_queue;
+}
+
+/**
+ * Query for FOIA requests whose status is set to "To Be Scanned".
+ */
```
dangur: Should this have a description of what this function returns?

docroot/modules/custom/foia_request/foia_request.drush.inc
```diff
@@ -0,0 +1,158 @@
...
+function foia_request_check_file_fields($request) {
+  $attachment_fields = &drupal_static(__FUNCTION__);
+  $webform_submission = WebformSubmission::load($request->sid);
+
+  if (isset($attachment_fields[$request->webform_id])) {
+    $file_fields = $attachment_fields[$request->webform_id];
+  }
+  else {
+    $webform = $webform_submission->getWebform();
+    $file_field_lookup = \Drupal::service('foia_webform.file_field_lookup_service');
```
dangur: Should we use one of the preexisting getFileAttachmentElementsOnWebform protected functions from the FoiaSubmissionServiceApi or FoiaSubmissionQueueHandler classes instead of creating a new lookup service for fields?

docroot/modules/custom/foia_webform/src/FileFieldLookupService.php
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * Class FileFieldLookupService.
+ */
+class FileFieldLookupService {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getFileAttachmentElementsOnWebform(WebformInterface $webform) {
```
dangur: If we want to keep this, it seems better if it were protected rather than public.

docroot/modules/custom/foia_webform/src/FileFieldLookupServiceInterface.php
```diff
@@ -0,0 +1,24 @@
...
+   * Gets the machine names of all file attachment elements on the webform.
+   *
+   * @param \Drupal\webform\WebformInterface $webform
+   *   The webform being submitted against.
+   *
+   * @return array
+   *   Returns an array of machine names of file attachment elements on the
+   *   webform being submitted against.
+   */
+  public function getFileAttachmentElementsOnWebform(WebformInterface $webform);
```
dangur: Similarly, if we want to keep this, it seems better if it were protected rather than public.

docroot/modules/custom/foia_request/foia_request.drush.inc
Tom-Camp: The intent is to remove duplicate code and turn it into a service. We should refactor and remove the methods from the other classes and use the service.

docroot/modules/custom/foia_webform/src/FileFieldLookupServiceInterface.php
Tom-Camp: That function needs to be called from other classes, so it needs to be public, unless we add a different public method that calls that method.

Tom-Camp:
The intent with making the `file_field_lookup_service` was to reduce the duplicate code that is currently being used. We were doing exactly the same thing in 3 different places so a service seemed like the best bet. This PR does not handle refactoring the other places where the duplicate code lives.

ba66e77:
@Tom-Camp, can you create an enhancement ticket to refactor the "other places where the duplicate code lives"? That way we at least have record/notes of what needs to be cleaned up even if it isn't priority at this moment. 

//cc @malikkotob

PR #124, opened 2017-12-19
DOJ-457: Update submission queue handler

Tom-Camp:


docroot/modules/custom/foia_webform/src/Plugin/WebformHandler/FoiaSubmissionQueueHandler.php
```diff
@@ -32,7 +33,11 @@ public function postSave(WebformSubmissionInterface $webformSubmission, $update
       $componentAssociatedToWebform = $this->getComponentAssociatedToWebform($webformSubmission);
       if ($componentAssociatedToWebform) {
         $foiaRequest = $this->createFoiaRequest($webformSubmission, $componentAssociatedToWebform);
-        $this->queueFoiaRequest($foiaRequest);
+      }
+
+      if ($foiaRequest && $foiaRequest->getRequestStatus() === FoiaRequestInterface::STATUS_QUEUED) {
+        $queuer = new FoiaSubmissionQueueingService();
+        $queuer->addRequestToQueue($foiaRequest);
```
malikkotob: I'm thinking these three lines of code can actually exist within the `queueFoiaRequest` method, so it would be:
```
if ($foiaRequest->getRequestStatus() === FoiaRequestInterface::STATUS_QUEUED) {
  $queuer = new FoiaSubmissionQueueingService();
  $queuer->addRequestToQueue($foiaRequest);
}
else {
  return FALSE;
}
```

or something like that. That way the code within the `postSave` method still reads as "create a request, queue a request"

docroot/modules/custom/foia_webform/src/FoiaSubmissionQueueingService.php
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * Class FoiaSubmissionQueueingService.
+ */
+class FoiaSubmissionQueueingService implements FoiaSubmissionQueueingServiceInterface {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function addRequestToQueue(FoiaRequestInterface $foiaRequest) {
```
malikkotob: any reason not to simply name this `queueRequest` or `queue`?

docroot/modules/custom/foia_webform/src/FoiaSubmissionQueueingService.php
```diff
@@ -0,0 +1,36 @@
+<?php
+
+namespace Drupal\foia_webform;
+
+use Drupal\foia_request\Entity\FoiaRequestInterface;
+
+/**
+ * Class FoiaSubmissionQueueingService.
+ */
+class FoiaSubmissionQueueingService implements FoiaSubmissionQueueingServiceInterface {
```
malikkotob: Since the method below queues a `foiaRequest`, maybe we can call this `FoiaRequestQueuer` or `FoiaRequestQueuingService`

docroot/modules/custom/foia_webform/src/FoiaSubmissionQueueingService.php
```diff
@@ -0,0 +1,36 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public function addRequestToQueue(FoiaRequestInterface $foiaRequest) {
+    /** @var \Drupal\Core\Queue\QueueFactory $queueFactory */
+    $queueFactory = \Drupal::service('queue');
+
+    // @var QueueInterface $queue
+    $foiaSubmissionsQueue = $queueFactory->get('foia_submissions');
```
malikkotob: I see why you used the word `submission` in the class title now.

docroot/modules/custom/foia_webform/src/FoiaSubmissionQueueingService.php
```diff
@@ -0,0 +1,36 @@
...
+ * Class FoiaSubmissionQueueingService.
+ */
+class FoiaSubmissionQueueingService implements FoiaSubmissionQueueingServiceInterface {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function addRequestToQueue(FoiaRequestInterface $foiaRequest) {
+    /** @var \Drupal\Core\Queue\QueueFactory $queueFactory */
+    $queueFactory = \Drupal::service('queue');
```
malikkotob: We should probably make this class an actual service and inject these items using dependency injection.

docroot/modules/custom/foia_webform/src/FoiaSubmissionQueueingServiceInterface.php
```diff
@@ -0,0 +1,20 @@
...
+ */
+interface FoiaSubmissionQueueingServiceInterface {
+
+  /**
+   * Adds a FOIA request to the foia_submissions queue.
+   *
+   * @param \Drupal\foia_request\Entity\FoiaRequestInterface $foiaRequest
+   *   The FOIA Request to queue for later processing.
+   */
+  public function addRequestToQueue(FoiaRequestInterface $foiaRequest);
```
malikkotob: Nice job using an interface here.

docroot/modules/custom/foia_webform/src/FoiaSubmissionQueueingService.php
```diff
@@ -0,0 +1,62 @@
...
+  public function __construct(QueueFactory $queueFactory, LoggerInterface $logger) {
+    $this->queueFactory = $queueFactory;
+    $this->logger = $logger;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function enqueue(FoiaRequestInterface $foiaRequest) {
+    // @var QueueInterface $queue
```
malikkotob: I think the right syntax for type hinting is:
```
/** @var \Drupal\Core\Queue\QueueInterface $foiaSubmissionsQueue */
```
With that being said, I don't think we actually need the type hint, since the `$queueFactory` explicitly returns an object of type `QueueInterface`.

docroot/modules/custom/foia_webform/tests/src/Kernel/FoiaSubmissionQueueHandlerTest.php
```diff
@@ -189,12 +190,11 @@ public function testFoiaRequestAttachmentPendingScan() {
...
 
-    $queuedSubmission = $this->foiaSubmissionsQueue->claimItem()->data;
+    $queueCount = $this->foiaSubmissionsQueue->numberOfItems();
 
-    $this->assertNotEmpty($queuedSubmission, "Expected a FOIA request ID to be queued, but nothing was found in the queue.");
-    $this->assertEquals('1', $queuedSubmission->id, "Queued FOIA Request ID does not match expected.");
+    $this->assertEquals($originalQueueCount, $queueCount, "Expect that the submission with attachment would not be queued, it was.");
 
-    $foiaRequest = FoiaRequest::load($queuedSubmission->id);
+    $foiaRequest = FoiaRequest::load($this->webformSubmission->id());
```
malikkotob: I think this is working because the id of the webform submission happens to be the same as the id of the foia request. In reality though, the FOIA request will not have the same ID as the webform submission. Instead, we could either load the request by writing a query to find it based on the submission ID. Or we could use Drupal's entity storage and do a `loadByProperties` to load the foia request based on the webform submission ID.

PR #137, opened 2018-01-23
Doj vm setup

brockfanning:
These are some initial steps towards a local development box that will work in the locked-down DOJ network/laptops.

Important note: This changes the domain of the BLT box, including the Drush alias, so existing VMs may need to be re-provisioned or re-built.

### DIFF
--- a/blt/project.yml
+++ b/blt/project.yml
@@ -6,7 +6,7 @@ project:
     name: lightning
   local:
     protocol: http
-    hostname: 'local.${project.machine_name}.gov'
+    hostname: 'local-api.foia.doj.gov'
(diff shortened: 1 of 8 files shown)


brockfanning:
@malikkotob @ba66e77 Just wanted to double-check - is it OK to merge this? I think it will probably cause your local drush alias to stop working, because the local URL will be expected to change.

malikkotob:
@brockfanning, I finally got a chance to pull these changes in locally and do some brief regression testing against my machine. Everything worked just fine and automated tests passed (i.e. running `blt tests`). Nicely done!

PR #140, opened 2018-02-01
DOJ-242: API improvement: add field in agency component to denote it is a centralized agency

dangur:


docroot/modules/custom/foia_ui/foia_ui.install
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * Set Agency Component Centralized field.
+ */
+function foia_ui_update_8001() {
+
+  // Get all components.
+  $bundle = 'agency_component';
+  $query = \Drupal::entityQuery('node');
+  $query->condition('type', $bundle);
+  $componentEntityIds = $query->execute();
```
malikkotob: Hmm, I wonder if it would be more efficient to start the search at the agency level.
e.g.
1) Query to "find all agencies associated with more than 1 component", e.g:
```
select nfa.`field_agency_target_id` from node__field_agency nfa group by nfa.`field_agency_target_id` having COUNT(nfa.`field_agency_target_id`) > 1;
```
This will return an array of term IDs

2) For each agency term ID, query to "find all components associated with that agency", e.g:
```
select nfa.entity_id from node__field_agency nfa where nfa.`field_agency_target_id` = 1251;
```
This will return an array of nids of components for each of the agencies

3) For each component, update the `is_centralized` field to false

Then, for the centralized agencies, we'd do the same thing but the first query would be modified to find the agencies when the count = 1.

I think this reduces the number of queries to: the total number of agencies + 2 (~119). The existing approach runs a query for each component + the first query to find all components (>400 queries).

PR #146, opened 2018-02-13
Additional tweaks to DOJ-specific VM setup

brockfanning:


brockfanning:
@malikkotob This has a few clarifications and small fixes - any objection to merging this, if it doesn't interfere with code-freeze concerns?

PR #150, opened 2018-03-01
DOJ-149: Patch jsonapi_extras for performance problem.

brockfanning:


brockfanning:
Closing this since the patch was merged into the latest release.

brockfanning:
Re-opening this because of Composer complications. We may need to apply this patch for now.

PR #166, opened 2018-04-26
DOJ-591: Allow CSV Export of Request Tracking Info.

cnovick-doj:


cnovick-doj:
@brockfanning This is ready for you to review.

brockfanning:
@cnovick-doj I see you added a "Created since" filter. That seems useful, but I think if you're going to include it, it needs help text about what format the date should be in (YYYY-MM-DD).

cnovick-doj:
Will add that to the field description.

cnovick-doj:
@brockfanning Date format has been added to exposed filter

PR #188, opened 2018-12-11
DOJ-829306: Allow dollar signs in all webforms on 'fee' input.

brockfanning:


brockfanning:
Leaving this PR open until release time, in case webform changes need to be rebased into it.

PR #201, opened 2019-04-30
DOJ-879160: Adds a time and date stamps to the FOIA requests sent fro…

Amstercad:
At the time of FOIA request submission, this code adds a time and date stamp to each FOIA Request sent from the National Portal.


brockfanning:
@Amstercad I would recommend getting the latest "develop" code and creating a new PR. The "develop" branch has been "rewound" a bit to omit some recent unintentional commits. Also, make sure to create your PR against the "develop" branch. (The goal is to merge your changes into the develop branch.)

Some comments on the approach:
1. Rather than setting the timezone for all future dates, let's just set the timezone only for this particular variable. See the alternative approach [here](https://stackoverflow.com/a/19617551).
2. I believe the valid timezone string is "America/New_York".

PR #217, opened 2019-07-31
Updating SimpleSAMLPHP library.

ba66e77:
Updating SimpleSAMLPHP library to latest version.


ba66e77:
Note that after updating the vendor library you need to run `blt source:build:simplesamlphp-config` to copy the configs into the vendor dir

PR #230, opened 2019-10-09
FOIA-178: Increase memory limit on Annual Reports. (#224)

bemarlan:
* FOIA-178: Increase memory limit on Annual Report form submissions.

* FOIA-178: Add config and settings form for config options to set memory limit value and debug.

### DIFF
--- a/config/default/block.block.bartik_content.yml
+++ b/config/default/block.block.bartik_content.yml
@@ -11,7 +11,7 @@ _core:
 id: bartik_content
 theme: bartik
 region: content
-weight: -2
+weight: -3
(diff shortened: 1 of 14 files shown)


ba66e77:
Note that since the module is already installed, the config in `config/install` included with it _will not_ be automatically applied.

bemarlan:
**FOIA-178**
As Barrett mentioned, when releasing to each environment, we will need to set configuration manually. I will send instructions in email.

**FOIA-187**
Once this gets to production, we will want to remove the configuration that was added in this PR from the secret settings file on production.

PR #260, opened 2020-01-02
FOIA-286: Revert strict comparisons in foia_ui validation.

bemarlan:


brockfanning:
@bemarlan Was the composer.lock change supposed to be part of this? I can't see any meaningful change in there, but I may be missing something.

PR #271, opened 2020-03-16
Revert "FOIA-000: Merge Acquia's development work into DOJ's develop."

brockfanning:
Reverts usdoj/foia-api#266


brockfanning:
@jesseday I'm considering reverting this chunk of code temporarily, so that we can get the bulk import onto production without causing issues. Do you think the bulk import will work without this chunk of code?

PR #315, opened 2021-04-27
FOIA-1189304: Security Update hotfix.

jhuang0427:


brockfanning:
@jhuang0427 It looks like there is an updated patch: https://www.drupal.org/files/issues/2021-04-06/corrupted_data_paragraphs-3060223-46.patch

I found that on the issue page: https://www.drupal.org/node/3060223

Did you try using this one?

jhuang0427:
I have committed to PR with alternative on drupal/entity_clone

composer.json
```diff
@@ -125,6 +125,9 @@
             },
             "drupal/entity_clone": {
```
brockfanning, line 126: @jhuang0427 We only need one "drupal/entity_clone" entry - you can remove the old one and just have the new one.

jhuang0427:
have removed older one.

PR #326, opened 2021-10-08
FE2-20: setup form mode

rgristroph:
This has the yaml files for each form display mode, but the fields are not appropriately hidden for each one yet.


rgristroph:
This was made to the wrong repo, closing here, it will appear later as part of a larger PR.

PR #333, opened 2021-11-03
FOIA-106: Changed the status type.

sundarcreosen:


config/default/views.view.foia_requests.yml
```diff
@@ -892,7 +892,7 @@ display:
                 title: 'Queued for submission'
                 operator: '='
                 value:
-                  value: '0'
+                  value: '3'
```
brockfanning, line 895: Do you know why this got changed to 3?

brockfanning (changes requested):
@sundarcreosen Just had one question about that value changing to 3.

Also I think this also needs to be updated: https://github.com/usdoj/foia-api/blob/develop/docroot/modules/custom/foia_request/src/Entity/FoiaRequest.php#L225

config/default/views.view.foia_requests.yml
sundarcreosen: After updating the email transit to delete after 7 days when I save the node it doesn't save it's asks for option value. In group 1 option value was empty I couldn't able to save it. There were already option value was 1,2,4 so I updated 3 and saved it.

brockfanning: @sundarcreosen I have tried it locally and I see what you're saying. However the correct value is 0. Please change it to 0 instead of 3.

PR #343, opened 2021-11-22
FOIA-922: Updated the permission.

sundarcreosen:


sundarcreosen:
Updated the permission. Please verify @brockfanning

PR #346, opened 2021-12-01
FOIA-249: Make address field required - (to prevent FOIA contacts dow…

sundarcreosen:
…nload failures).


sundarcreosen:
@brockfanning Please verify the change.

PR #460, opened 2022-12-29
FOIA-68: Convert all front-stage form validation to back-stage.

ychen-doj:


docroot/modules/custom/foia_api/src/Plugin/rest/resource/WebformSubmissionResource.php
```diff
@@ -136,6 +136,26 @@ public function post($data) {
...
+      $addressZipPostalCode = $data['address_zip_postal_code'] ?? '';
+      $country = $data['address_country'] ?? '';
+      $mailingAddress = (!$addressLine1 || !$addressCity || !$addressStateProvince || !$addressZipPostalCode || !$country) ? '' : 'mailling address';
+    }
+
+    if (!$mailingAddress && !$email && !$phoneNumber) {
+      $statusCode = 422;
+      $message = t("At least Email, Mailing address, or Phone are required.");
+      $this->logSubmission($statusCode, $message);
+      return new ModifiedResourceResponse(['errors' => $message], $statusCode);
```
brockfanning, line 157: I believe that the 'errors' key should be pointing to an object of field names to error messages - not just a string. You can see an example returned if you submit a request via API with a missing required field. I think in our case we want it to be the same message all three fields -- something like:
```
$message = 'In order to submit your request, you must provide at least one of the following: email address, mailing address, or phone number.';
$errors = [
  'email' => $message,
  'phone' => $message,
  'address_line1' => $message,
];
```

brockfanning (changes requested):
@ychen-doj A few specific comments, and also a more general one: I think this needs to be moved to a different place in the file, such that it can use the same conditional as the current "Validate submission" section.

brockfanning (approved):

PR #464, opened 2023-01-12
Revert "FOIA-363: request failed submissions handler."

brockfanning:
Reverts usdoj/foia-api#453


brockfanning:
Reverting this because this PHP error:

```
Error: Call to a member function getData() on null in Drupal\foia_webform\Plugin\QueueWorker\FoiaSubmissionQueueWorker->handleFailedSubmission() (line 204 of /mnt/www/html/foiauat/docroot/modules/custom/foia_webform/src/Plugin/QueueWorker/FoiaSubmissionQueueWorker.php).
```

PR #467, opened 2023-01-13
149880 request failed submitssion handler

ychen-doj:


docroot/modules/custom/foia_webform/src/Plugin/QueueWorker/FoiaSubmissionQueueWorker.php
```diff
@@ -200,12 +199,16 @@ protected function handleFailedSubmission(FoiaRequestInterface $foiaRequest, arr
...
+      // Get webform submission id.
+      $subMissionId = $foiaRequest->get('field_webform_submission_id')->getString();
+      // Load webform submission data.
+      $webform_submissions = \Drupal::entityTypeManager()->getStorage('webform_submission')->load($subMissionId);
       $submissionsData = $webform_submissions->getData();
       $email = $submissionsData['email'];
       $date = date('m/d/Y h:i:s a', $foiaRequest->get('field_submission_time')->getString());
       $errormsg = sprintf('FOIA request failed too many times. Agency: %s, Component: %s, Date: %s, Email: %s', $agencyComponentName, $agencyName, $date, (empty($email) ? '(not provided)' : $email));
       // Log a unique message that this happened.
+      $foiaRequest->set('field_error_message', $errormsg);
```
brockfanning, line 211: @ychen-doj Let's not add this here. The error message that is in the request needs to preserve whatever error message was originally there.

Otherwise this fix makes sense to me.

brockfanning (approved):

PR #493, opened 2023-03-14
FOIA-379: Bulk Publish for Annual Reports.

ychen-doj:


brockfanning:
@ychen-doj Let's add a test do this, now that we have the ability to do that. You could add one to the tests/behat/features/AnnualFOIAReportData.feature file, and it could be something simple like this, just to confirm that the option is there:

```
  @api @javascript
  Scenario: Agency Administrator see the option to bulk-publish annual reports
    Given annual_foia_report_data content:
      | field_agency | field_foia_annual_report_yr | moderation_state |
      | Federal Testing Agency | 2023 | cleared |
    And I am logged in as a user with the 'Agency Administrator' role
    And I am on "/admin/content/reports"
    And I select "Publish annual reports" from "Action"
```

This is basically just asserting the existence of the "Publish annual reports" option in the dropdown. If that option does not exist, the last step will fail.

brockfanning (approved):

PR #500, opened 2023-03-29
Revamp of agency/component widgets for reports

brockfanning:
This is an attempt to revamp the authoring workflow for the annual and quarterly reports. Currently the author needs to enter a year and agency, then click outside somewhere, and wait for the page to refresh, and then they can continue. This workflow is fine but is very difficult to test, and requires a lot of custom code. This attempt makes these changes:
1. The "Agency" field is a select list instead of autocomplete
2. The "Components" field is dynamic and the available checkboxes change when an "Agency" is selected (functionality provided by the contrib "Business Rules" module)
3. When the report is first being created, most of the fields/sections are hidden. This is intended to force the user to click "Save and continue".

The hope is that this workflow is simpler for automated testing and also removes a lot of custom code. @ameshkin @ychen-doj If time allows I'd appreciate if you could test it out, try to break it, and give some feedback.


ychen-doj:
I have walk through the three steps you mentioned above, everything good. Remember in thicket: WS-374 TASK0166139, we have set field "field_agency_components" as required field, does it will aganst the click "Save and continue" When the report is first being created?

ameshkin (approved):

brockfanning:
Thanks @ychen-doj good point. I think that will work out, because the author will need to select the checkboxes before clicking "Save and continue".

PR #534, opened 2023-08-01
Adds Agency Component views CSV export.

mattdanger:
Adds a new views CSV export that includes Agency Component information with these columns:

- Agency Component
- Agency Component Description
- Commonly Requested Records
- Agency
- Mission

The export is accessible to all users with the "View published content" permission, which is essentially everyone including non authenticated visitors.

Path: `export/agency-component-export.csv`


mattdanger:
I spoke with Jay @ Polydelta this morning. They have built a client that can consume the Agency Component and Agency data directly from Drupal. So, this Drupal views export is no longer needed. Closing!

PR #535, opened 2023-08-02
Added custom FOIA Wizard module

mattdanger:
Adds the `foia_wizard` custom module which creates a new Drupal CMS settings page for the config.

Wizard API endpoint: `/api/foia_wizard`

Path: `/admin/config/system/foia-wizard`


mattdanger:
Closing - we'll need to re-roll this with a version that does not contain our buildkitex workflow changes.

PR #571, opened 2023-10-24
FOIA Wizard: Make messages 48 through 51 editable

mattdanger:


mattdanger:
@brockfanning sure thing. I am going to close this PR for now so we can include few other small tweaks. New PR coming shortly.

PR #583, opened 2023-11-24
Remove references to drush9 from site aliases.

brockfanning:


brockfanning:
@catchenry I couldn't get remote drush commands to work in Cloud IDE unless I remove these lines. Does this look OK?

PR #600, opened 2024-02-06
FOIA-0000: Fixing cache issue for fiscal years and wizard.

ameshkin:


brockfanning:
@ameshkin Also this is probably a good chance to add a quick Behat test to assert that /api/annual_foia_report/fiscal_years gets a 200 response.

brockfanning:
@ameshkin And while you're at it, a similar test for: api/foia_wizard (asserting that it gets a 200 response)

ameshkin:
The PR for the behat tests is here
https://github.com/usdoj/foia-api/pull/602

PR #632, opened 2024-07-17
Updating drupal to 10.3.1

ameshkin:
The rules module was removed to get the site working with 10.3.1. However, this module is used to send an email when a new user signs up to the site.  I'm creating this draft PR for now to easily compare to develop.  Once there is a patch or update to the rules module, I'll update the branch and all commits will be squashed into one.

**FILE FOR SENDING EMAIL:**
https://github.com/usdoj/foia-api/blob/develop/config/default/rules.reaction.new_user_notification.yml

**ISSUE:**
https://www.drupal.org/project/rules/issues/3454056


ameshkin:
Added config_split module which fixed github test error.  Still not ready to merge without a fix for drupal/rules.

ameshkin:
Added rules back to composer.json file.

brockfanning (approved):

PR #644, opened 2024-08-23
FOIA-458: Upgrading autologout to version 2 and also checking alterna…

ameshkin:
…tive logout.


ameshkin:
I manually added js_cookie due to an issue with autologout that has still not been fixed. There is a patch but I believe the module contributors will fix it soon.

https://www.drupal.org/project/autologout/issues/3449275

@brockfanning

brockfanning (approved):

PR #751, opened 2026-03-23
FOIA-548: COOP header on backend.

brockfanning:


rgristroph-usdoj (approved):
Looks good.  This can also be done in the seckit settings in Drupal, I'm not sure there is a big advantage either way

